PR #2, opened 2020-12-07
Config service v1 Implementation

GurtejSohi:


config-service-impl/src/main/java/org/hypertrace/config/service/ConfigResource.java
```diff
@@ -0,0 +1,12 @@
+package org.hypertrace.config.service;
+
+import lombok.Data;
+
+@Data
```
aaron-steinfeld, line 5: Consider making this immutable with `@Value` - I didn't notice anywhere where we were mutating it, which is good.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -0,0 +1,139 @@
+package org.hypertrace.config.service;
+
+import com.google.protobuf.Struct;
+import com.google.protobuf.Value;
+import io.grpc.stub.StreamObserver;
+import java.util.LinkedHashMap;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.StringUtils;
```
aaron-steinfeld, line 8: Given that we're already using guava in this project as part of grpc, is there a compelling reason to bring in a second util lib?

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
```diff
@@ -0,0 +1,99 @@
...
+import com.google.protobuf.util.JsonFormat;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.core.documentstore.Document;
+
+import java.io.IOException;
+
+@Data
```
aaron-steinfeld, line 25: Same comment about mutability

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocumentKey.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Key for the Config Document
+ */
+public class ConfigDocumentKey implements Key {
+
+  private static final String SEPARATOR = ":";
+
+  private final ConfigResource configResource;
+  private final long configVersion;
```
aaron-steinfeld, line 14: Should the version be part of the resource?

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Get the config for the specified resource with the specified version(optional). If
+   * configVersion is empty, get the config for the latest version.
+   *
+   * @param configResource
+   * @param configVersion
+   * @return
+   */
+  GetConfigResponse getConfig(ConfigResource configResource, Optional<Long> configVersion)
```
aaron-steinfeld, line 40: generally we don't want to make optional part of an API - instead, overload getConfig for a version with long and version without. The reason here is that using an Optional in an api like this provides multiple different ways of omitting a value - a null or an empty optional, leaving ambiguity in the API and adding more work internally to account for more input states.

config-service/build.gradle.kts
```diff
@@ -10,8 +10,34 @@ plugins {
...
+    }
+    runtimeOnly("io.netty:netty-codec-http2:4.1.53.Final") {
+      because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1020439")
+    }
+    runtimeOnly("io.netty:netty-handler-proxy:4.1.53.Final") {
+      because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1020439")
+    }
+  }
+
+  implementation("org.hypertrace.core.grpcutils:grpc-server-utils:0.2.0")
```
aaron-steinfeld, line 26: There's some old versions in here. Since it's a new service, can you audit them and bring them all up to latest.

config-service/build.gradle.kts
```diff
@@ -10,8 +10,34 @@ plugins {
...
 }
 
 application {
-  mainClassName = "org.hypertrace.config.service.MyApplication"
+  mainClassName = "org.hypertrace.core.serviceframework.PlatformServiceLauncher"
 }
+
+// Config for gw run to be able to run this locally. Just execute gw run here on Intellij or on the console.
+tasks.run<JavaExec>  {
+  jvmArgs = listOf("-Dbootstrap.config.uri=file:${projectDir}/src/main/resources/configs", "-Dservice.name=${project.name}")
```
aaron-steinfeld, line 42: bootstrap config URI not needed, this will be picked up by default (I believe service.name is still needed though)

config-service/build.gradle.kts
```diff
@@ -10,8 +10,34 @@ plugins {
...
+  runtimeOnly("io.grpc:grpc-netty:1.30.2")
+  implementation("com.typesafe:config:1.4.0")
+
+  implementation("org.slf4j:slf4j-api:1.7.30")
+  runtimeOnly("org.apache.logging.log4j:log4j-slf4j-impl:2.13.3")
 }
 
 application {
-  mainClassName = "org.hypertrace.config.service.MyApplication"
+  mainClassName = "org.hypertrace.core.serviceframework.PlatformServiceLauncher"
```
aaron-steinfeld, line 37: this is deprecated in favor of the lazy property -
```suggestion
  mainClass.set("org.hypertrace.core.serviceframework.PlatformServiceLauncher")
```

config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @Override
+  public String getServiceName() {
+    return serviceName;
+  }
+
+  private ConfigStore getConfigStore(Config config) {
+    try {
+      String providerClassName = config.getString(CONFIG_STORE_CONFIG);
```
aaron-steinfeld, line 87: Unless we're planning on making runtime-pluggable modules for new config stores, what's the value in doing this via class loading - it's an easy area to slip up when renaming + refactoring.

helm/values.yaml
```diff
@@ -1 +1,96 @@
...
+  periodSeconds: 5
+
+resources:
+  # We usually recommend not to specify default resources and to leave this as a conscious
+  # choice for the user. This also increases chances charts run on environments with little
+  # resources, such as Minikube. If you do want to specify resources, uncomment the following
+  # lines, adjust them as necessary, and remove the curly braces after 'resources:'.
+   requests:
+     cpu: 0.1
+     memory: 1024Mi
```
aaron-steinfeld, line 51: Does this need a full gb of memory? Seems pretty resource light by default.

helm/templates/deployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+              containerPort: {{ .Values.containerAdminPort }}
+              protocol: TCP
+          env:
+            - name: SERVICE_NAME
+              value: "{{ .Chart.Name }}"
+            - name: BOOTSTRAP_CONFIG_URI
+              value: "file:///app/resources/configs"
+            - name: LOG4J_CONFIGURATION_FILE
+              value: "/var/{{ .Chart.Name }}/log/log4j2.properties"
+            - name: JAVA_TOOL_OPTIONS
```
aaron-steinfeld, line 67: Can we use `JAVA_OPTS`

helm/templates/deployment.yaml
```diff
@@ -0,0 +1,87 @@
...
+          ports:
+            - name: grpc-port
+              containerPort: {{ .Values.containerPort }}
+              protocol: TCP
+            - name: admin-port
+              containerPort: {{ .Values.containerAdminPort }}
+              protocol: TCP
+          env:
+            - name: SERVICE_NAME
+              value: "{{ .Chart.Name }}"
```
aaron-steinfeld, line 62: pretty sure the dockerfile already defines this

helm/templates/config-service-config.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ .Values.configServiceConfig.name }}
+  labels:
+    release: {{ .Release.Name }}
+data:
+  application.conf: |-
+    document.store {
```
aaron-steinfeld, line 9: are we not exposing the other properties of the conf? Specifically, the ports, since those are already parameterized elsewhere.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -0,0 +1,139 @@
...
+      responseObserver.onError(e);
+    }
+  }
+
+  @Override
+  public void getConfig(GetConfigRequest request,
+      StreamObserver<GetConfigResponse> responseObserver) {
+    try {
+      GetConfigResponse response;
+      if (request.getConfigVersion() != 0) {  // if configVersion is present
```
aaron-steinfeld, line 50: Similar to what was done on upsert, I'd consider moving the bulk of this into a separate class so we don't end up with one giant file containing the whole service.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        String context = null;
```
aaron-steinfeld, line 51: Avoiding any nulls by designing apis to account for the null case would be better if possible (i.e. getConfigResourceWithoutContext or similar)

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -0,0 +1,139 @@
...
+        // get the config version corresponding to the most specific context(last element in the contexts list)
+        if (!request.getContextsList().isEmpty()) {
+          context = request.getContexts(request.getContextsCount() - 1);
```
aaron-steinfeld, line 54: This is an interesting question with unclear behavior - if I ask for a version and a list of scopes, we're only returning the last. That's unexpected to me, and I think indicates that the version doesn't belong in the same api. My suggestion would be to remove version from the get API and when/if needed, introduce a new API for a versioned request that does not support merging/multiple contexts.

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
aaron-steinfeld: I also think it's a bit strange to mix concerns here by returning a GetConfigResponse - the data store is now coupled to the external API. Would it make sense to return a value proto (or some new object if there's more data needed)

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -0,0 +1,139 @@
...
+            merge(defaultConfigMap.get(entry.getKey()), entry.getValue()));
+      }
+      Struct struct = Struct.newBuilder().putAllFields(resultConfigMap).build();
+      return Value.newBuilder().setStructValue(struct).build();
+    } else {
+      return overridingConfig;
+    }
+  }
+
+  // TODO : get the userId from the context
```
aaron-steinfeld, line 135: I'd recommend omitting this whole concept we have a use case and need. userId itself is ambiguous, as it's unclear what user it represents. Is it the author, the last editor, the user it applies to? All of those are potentially things we might want to track. Also, many configs may driven by the system.

aaron-steinfeld (commented):
Please also add a README

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigResource.java
GurtejSohi: This was immutable previously also(declared as `final`). But still replaced it with `@Value` to make it a bit compact.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: removed the Apache commons lib

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
GurtejSohi: Updated

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocumentKey.java
GurtejSohi: I like to think of it as each resource can have multiple versions of configs associated with it. This way the APIs are also cleaner as some APIs only require the resource and not the version.

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
GurtejSohi: Totally makes sense. Updated it.

config-service/build.gradle.kts
GurtejSohi: Updated

config-service/build.gradle.kts
GurtejSohi: Removed it

config-service/build.gradle.kts
GurtejSohi: Updated

config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
GurtejSohi: Sure, initially, we were planning to try with multiple stores which is why we had this. But as you pointed out it's no longer needed. Updated it.

helm/values.yaml
GurtejSohi: reduced it to 512 MiB

helm/templates/deployment.yaml
GurtejSohi: done

helm/templates/deployment.yaml
GurtejSohi: removed it

helm/templates/config-service-config.yaml
GurtejSohi: We need to do it. Updated.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: Moved a couple of methods to a `Utils` class

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: done

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: Removed version from `getConfig` API but still keeping it in store, so that we can easily support the new API(for getting a specific version of config) easily.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: `UserId` here represents the user who upserted the config - author/creator and last editor are same as once a config version is inserted into the store, it is never updated/editted. We always create a new version of the config.
I've just kept it here as a placeholder(so that the store can work with it) for now and is not exposed via the API. I think if we need to expose it via the API in future, we can document it to remove ambiguity, right?

config-service-impl/src/main/java/org/hypertrace/config/service/Utils.java
```diff
@@ -0,0 +1,46 @@
...
+
+public class Utils {
+
+  public static final String DEFAULT_CONTEXT = "DEFAULT-CONTEXT";
+
+  private Utils() {
+    // to prevent instantiation
+  }
+
+  public static Value merge(Value defaultConfig, Value overridingConfig) {
```
pavan-traceable, line 17: Please add testcases to this

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigResource.java
```diff
@@ -0,0 +1,12 @@
+package org.hypertrace.config.service;
+
+import lombok.Value;
+
+@Value
+public class ConfigResource {
```
pavan-traceable, line 6: doc pls

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -0,0 +1,99 @@
...
+import org.hypertrace.config.service.v1.GetConfigRequest;
+import org.hypertrace.config.service.v1.GetConfigResponse;
+import org.hypertrace.config.service.v1.UpsertConfigRequest;
+import org.hypertrace.config.service.v1.UpsertConfigResponse;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+import java.util.Optional;
+
+@Slf4j
+public class ConfigServiceGrpcImpl extends ConfigServiceGrpc.ConfigServiceImplBase {
```
pavan-traceable, line 21: Are you working on  unit tests ?

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aaron-steinfeld: Fair on edit vs author. I don't believe we use userId anywhere in the platform other than feature flags right now, so we may want to discuss how we want to expose it when we're ready for this. I'd rather not require services to have to start working with the JWT directly.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aaron-steinfeld: Might consider making this a constructor provided class (even if not using DI), to make it easier to test, since statics are generally a pain.

config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
aaron-steinfeld: Sounds good - even with support for multiple stores, specifying the constructor FQN and using class loading is only really necessary if the the implementations are developed independently, and thus the loading code has no other way of getting the class. The way we're doing it, the classes are still in the same project, so a simple key could suffice.

helm/values.yaml
aaron-steinfeld: May want to consider going even lighter - I was thinking 128M since this doesn't deal with high concurrency or complex logic. It can always be increased in any prod deployment since these are variables, but we want HT as a whole platform to be default runnable on a laptop, so the full collection of services should be in the 2-6GB range.

aaron-steinfeld:
Will also want to resolve snyk issues before merging

helm/values.yaml
GurtejSohi: Makes sense. Reduced it to 128 MiB

config-service-impl/src/main/java/org/hypertrace/config/service/Utils.java
```diff
@@ -0,0 +1,46 @@
+package org.hypertrace.config.service;
+
+import com.google.common.base.Strings;
+import com.google.protobuf.Struct;
+import com.google.protobuf.Value;
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+public class Utils {
```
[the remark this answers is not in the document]
GurtejSohi: Updated

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
```diff
@@ -0,0 +1,113 @@
...
+import com.google.protobuf.Value;
+import com.google.protobuf.util.JsonFormat;
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.core.documentstore.Document;
+
+import java.io.IOException;
+
+@lombok.Value
+@Slf4j
+public class ConfigDocument implements Document {
```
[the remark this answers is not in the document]
GurtejSohi: Added javadocs

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: Added unit tests

GurtejSohi:
> Will also want to resolve snyk issues before merging

Have resolved most of them, except one. For that vulnerability(`io.netty:netty-handler@4.1.53.Final
`), no upgrade or patch is available.

aaron-steinfeld:
> > Will also want to resolve snyk issues before merging
> 
> Have resolved most of them, except one. For that vulnerability(`io.netty:netty-handler@4.1.53.Final `), no upgrade or patch is available.

See an example of a snyk ignore file for those cases (may be same vulnerability): https://github.com/hypertrace/hypertrace-core-graphql/blob/main/.snyk

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceUtils.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  /**
+   * Deep merge the specified {@link Value} configs with overridingConfig taking precedence over
+   * defaultConfig for the same keys.
+   * @param defaultConfig
+   * @param overridingConfig
+   * @return the resulting config obtained after merging defaultConfig and overridingConfig
+   */
+  public static Value merge(Value defaultConfig, Value overridingConfig) {
```
aaron-steinfeld, line 27: Please create unit tests for this merging code specifically, as this is a decent amount of logic.

config-service/src/integrationTest/README.md
```diff
@@ -0,0 +1,9 @@
+<h4>How do I run the integration tests locally?</h4>
+
+- Bring up a mongodb container locally using Docker or port forward to the mongo container in Kubernetes
```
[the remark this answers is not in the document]
GurtejSohi: Right. Updated it. 
I think I copied it from attribute service. So, I've raised a PR for updating the README there as well: https://github.com/hypertrace/attribute-service/pull/44

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceUtils.java
GurtejSohi: Done

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigResource.java
GurtejSohi: Added

config-service-impl/src/main/java/org/hypertrace/config/service/Utils.java
GurtejSohi: Added

aaron-steinfeld (approved):

PR #3, opened 2020-12-18
Fix NPE encountered with config value

GurtejSohi:


config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceUtils.java
```diff
@@ -59,4 +59,8 @@ public static Value merge(Value defaultConfig, Value overridingConfig) {
   public static String getActualContext(String rawContext) {
     return Strings.isNullOrEmpty(rawContext) ? DEFAULT_CONTEXT : rawContext;
   }
+
+  private static boolean isNull(Value value) {
+    return value == null || value.getKindCase() == Value.KindCase.NULL_VALUE;
```
aaron-steinfeld, line 64: It's fine to be defensive, but out of curiosity - is it ever possible for value to actually be `null`? A proto should never  return null, that's one of their design principles, so any nulls we're encountering would be from our own code and we should consider a bug. Given that, should we log an error in the `null` case and convert it to a `NULL_VALUE`?

GurtejSohi: Yeah, value should not be `null`. Updated it to log an error on seeing a `null` value and convert it to `NULL_VALUE` wherever required.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -57,7 +58,7 @@ public void getConfig(GetConfigRequest request,
         config = merge(config, contextSpecificConfig);
       }
       responseObserver.onNext(GetConfigResponse.newBuilder()
-          .setConfig(config)
+          .setConfig(nullSafeValue(config))
```
aaron-steinfeld, line 61: This is just being defensive right? Our implementation no longer returns a null?

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceUtils.java
```diff
@@ -59,4 +62,22 @@ public static Value merge(Value defaultConfig, Value overridingConfig) {
   public static String getActualContext(String rawContext) {
     return Strings.isNullOrEmpty(rawContext) ? DEFAULT_CONTEXT : rawContext;
   }
+
+  private static boolean isNull(Value value) {
+    if (value == null) {
+      log.error("NULL Value encountered. This is unexpected and indicates a BUG in code.",
+          new RuntimeException());
```
aaron-steinfeld, line 69: does the error logger require an exception to the message? Doesn't look like it adds any value

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: Yes, we're just being defensive here. The implementation no longer returns a null value.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceUtils.java
GurtejSohi: I just thought in case we hit this, the stack trace can help pin point the place where we're seeing a null value, which can probably help us debug quickly.

aaron-steinfeld (approved):

PR #5, opened 2021-01-05
Add APIs for deleteConfig and getAllConfigs and modify semantics of upsertConfig

GurtejSohi:
## Description
- Add API for getting all configs (i.e. across all the contexts) for the specified request
- Add API for deleting the config for the specified request
- Update semantics of upsert config API - It should merge the specified config with the existing config and upsert the resulting config into the store. Also it should return the config which is upserted.

### Testing
Added and updated unit tests and integration tests for the changes made. 

### Checklist:

### DIFF
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -87,7 +87,7 @@ message DeleteConfigRequest {
   // required - namespace with which the config resource is associated
   string resource_namespace = 2;
 
-  // required - for default context, use "DEFAULT-CONTEXT"
+  // required
   string context = 3;
 }
 
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -31,7 +31,8 @@ message UpsertConfigRequest {
   // required - config value to be upserted
   google.protobuf.Value config = 3;
 
-  // optional - only required if config applies to a specific context
+  // optional - only required if config applies to a specific context.
+  // If empty, specified config is associated with a default context.
   string context = 4;
 }
 
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -7,27 +7,89 @@ package org.hypertrace.config.service.v1;
 import "google/protobuf/struct.proto";
 
 service ConfigService {
+  // Merges the specified config with the existing config and upserts the
+  // resulting config into the store. Also returns the config which is upserted.
   rpc UpsertConfig(UpsertConfigRequest) returns (UpsertConfigResponse) {}
+
+  // Gets the config for the specified request
   rpc GetConfig(GetConfigRequest) returns (GetConfigResponse) {}
+
+  // Gets all the configs(i.e. across all the contexts) for the specified request
+  rpc GetAllConfigs(GetAllConfigsRequest) returns (GetAllConfigsResponse) {}
+
+  // Deletes the config for the specified request
+  rpc DeleteConfig(DeleteConfigRequest) returns (DeleteConfigResponse) {}
 }
 
 message UpsertConfigRequest {
-  string resource_name = 1;  // required
-  string resource_namespace = 2;  // required - namespace with which the config resource is associated
-  google.protobuf.Value config = 3; // required - config value to be upserted
-  string context = 4; // optional - only required if config applies to a specific context
+  // required - name of the resource associated with the config
+  string resource_name = 1;
+
+  // required - namespace with which the config resource is associated
+  string resource_namespace = 2;
+
+  // required - config value to be upserted
+  google.protobuf.Value config = 3;
+
+  // optional - only required if config applies to a specific context
+  string context = 4;
 }
 
 message UpsertConfigResponse {
-  int64 config_version = 1; // the version allocated to the newly inserted configuration
+  // config value upserted into the store
+  google.protobuf.Value config = 1;
 }
 
 message GetConfigRequest {
-  string resource_name = 1;  // required
-  string resource_namespace = 2;  // required - namespace with which the config resource is associated
-  repeated string contexts = 3; // optional; order matters - specify the contexts in the order of increasing priority. Example: ["service1", "api1"]
+  // required - name of the resource associated with the config
+  string resource_name = 1;
+
+  // required - namespace with which the config resource is associated
+  string resource_namespace = 2;
+
+  // optional; order matters - specify the contexts in the order of increasing
+  // priority. Example: ["service1", "api1"].
+  // Default context is implicitly appended to the beginning of the list.
+  repeated string contexts = 3;
 }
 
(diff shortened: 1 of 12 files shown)


config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -7,27 +7,90 @@ package org.hypertrace.config.service.v1;
...
+}
+
+message DeleteConfigRequest {
+  // required - name of the resource associated with the config
+  string resource_name = 1;
+
+  // required - namespace with which the config resource is associated
+  string resource_namespace = 2;
+
+  // required - for default context, use "DEFAULT-CONTEXT"
```
aaron-steinfeld, line 90: Could you remove the default context comment - that's an implementation detail - we explicitly decided not to support deleting this.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -67,6 +76,36 @@ public void getConfig(GetConfigRequest request,
...
+      log.error("Get all configs failed for request:{}", request, e);
+      responseObserver.onError(e);
+    }
+  }
+
+  @Override
+  public void deleteConfig(DeleteConfigRequest request,
+      StreamObserver<DeleteConfigResponse> responseObserver) {
+    try {
+      // write an empty config for the specified config resource. This maintains the versioning.
```
aaron-steinfeld, line 99: Hmmm - would we expect a deletion to remove historical versions too? Is there ever a way to delete those?

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -7,27 +7,90 @@ package org.hypertrace.config.service.v1;
...
+
+  // optional - only required if config applies to a specific context.
+  // If empty, specified config is associated with a default context.
+  string context = 4;
 }
 
 message UpsertConfigResponse {
-  int64 config_version = 1; // the version allocated to the newly inserted configuration
+  // config value upserted into the store
+  google.protobuf.Value config = 1;
```
aaron-steinfeld, line 41: Given that this isn't used yet, we can make the breaking change - but the fact that the build passes indicates we haven't set up breaking change detection as part of CI yet - can we please do that in a follow up (this PR is already doing too much)

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -34,10 +41,12 @@ public ConfigServiceGrpcImpl(ConfigStore configStore) {
   public void upsertConfig(UpsertConfigRequest request,
       StreamObserver<UpsertConfigResponse> responseObserver) {
     try {
-      long configVersion =
-          configStore.writeConfig(getConfigResource(request), getUserId(), request.getConfig());
+      ConfigResource configResource = getConfigResource(request);
+      Value existingConfig = configStore.getConfig(configResource);
```
aaron-steinfeld, line 45: Out of curiosity, does doc store itself support merging? Should it?

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -101,6 +107,19 @@ public Value getConfig(ConfigResource configResource) throws IOException {
     return config;
   }
 
+  @Override
+  public List<ContextSpecificConfig> getAllConfigs(String resourceName, String resourceNamespace,
+      String tenantId) throws IOException {
+    List<ContextSpecificConfig> contextSpecificConfigList = new ArrayList<>();
+    for (String context : getAllContexts(resourceName, resourceNamespace, tenantId)) {
```
aaron-steinfeld, line 114: Can we optimize this? We're doing N+1 separate DB queries here if I'm reading this right, getAll should be doable as a single query. Edit: is this what the todo below is referring to?

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
GurtejSohi: Removed it from the comments. But should we explicitly check if the context is set to `DEFAULT-CONTEXT`, then deny that request?

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: I think the main reason for keeping the historical versions is to be able to retrieve the old versions. Here also, if the client deletes the config either by mistake or for some other reason and wants to retrieve the older version, I think it should be able to do it (assuming we expose the API for retrieving the old versions going forward).
With the current scheme of things, there is no way to delete the old versions. We may want to put a cap on the no. of recent versions maintained, and probably clean up the old versions on the server side.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
GurtejSohi: Oh, yes. I had commented out the step for detecting breaking change in the first PR thinking that I would uncomment it in a subsequent PR but forgot to do so. Will definitely do it in a follow-up PR.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: Doc store supports updating a document but it doesn't support deep merging(the way we require it) of keys. Even if it did, I don't think we would be able to use it here as we're creating a new document for every new version.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
GurtejSohi: Updated it.
With the previous approach, I was trying to minimize the no. of docs processed but had to make N+1 queries, as you pointed out. But that would have been possible with the distinct clause only. Without distinct clause, we'd still have to process more documents(the way we're doing it now).
With the updated approach, we're making 1 query only but have to process all versions of a config resource even though we want the latest one.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
aaron-steinfeld: That's not a bad idea - up to you

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aaron-steinfeld: Maintaining is fine for now - once we have a better idea how versions will be used we can revisit the implementation since the API won't change.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aaron-steinfeld: true

aaron-steinfeld (approved):

PR #9, opened 2021-01-08
feat: add spaces config service

aaron-steinfeld:
Spaces config service allows for managing space configuration rules to be evaluated during the ingestion pipeline in the SpaceEnricher, and is used as part of hypertrace/hypertrace#142 . Management of these rules will be exposed externally via graphql.

By the time the tests were written on this one, the PR got pretty large - although not a significant amount of logic. If it makes the review easier, I can split a few things up.

### DIFF
--- a/config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
+++ b/config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
@@ -1,0 +1,24 @@
+package org.hypertrace.config.proto.converter;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Message;
+import com.google.protobuf.MessageOrBuilder;
+import com.google.protobuf.Value;
+import com.google.protobuf.util.JsonFormat;
+
+public class ConfigProtoConverter {
+  private static final JsonFormat.Printer JSON_PRINTER = JsonFormat.printer();
+  private static final JsonFormat.Parser JSON_PARSER = JsonFormat.parser().ignoringUnknownFields();
+
+  public static void mergeFromValue(Value value, Message.Builder builder)
+      throws InvalidProtocolBufferException {
+    JSON_PARSER.merge(JSON_PRINTER.print(value), builder);
+  }
+
+  public static Value convertToValue(MessageOrBuilder messageOrBuilder)
+      throws InvalidProtocolBufferException {
+    Value.Builder valueBuilder = Value.newBuilder();
+    JSON_PARSER.merge(JSON_PRINTER.print(messageOrBuilder), valueBuilder);
+    return valueBuilder.build();
+  }
+}
--- a/config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
+++ b/config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
@@ -1,24 +1,25 @@
 package org.hypertrace.config.service;
 
-import com.typesafe.config.Config;
+import static org.hypertrace.core.grpcutils.server.InterceptorUtil.wrapInterceptors;
+
+import io.grpc.ManagedChannel;
+import io.grpc.ManagedChannelBuilder;
 import io.grpc.Server;
 import io.grpc.ServerBuilder;
+import java.io.IOException;
 import org.hypertrace.config.service.store.ConfigStore;
 import org.hypertrace.config.service.store.DocumentConfigStore;
-import org.hypertrace.core.grpcutils.server.InterceptorUtil;
 import org.hypertrace.core.serviceframework.PlatformService;
 import org.hypertrace.core.serviceframework.config.ConfigClient;
+import org.hypertrace.space.config.service.SpacesConfigServiceImpl;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.io.IOException;
-
 public class ConfigService extends PlatformService {
 
-  private static final String SERVICE_NAME_CONFIG = "service.name";
   private static final String SERVICE_PORT_CONFIG = "service.port";
+  private static final String GENERIC_CONFIG_SERVICE_CONFIG = "generic.config.service";
   private static final Logger LOG = LoggerFactory.getLogger(ConfigService.class);
-  private String serviceName;
   private int serverPort;
   private Server configServer;
   private ConfigStore configStore;
@@ -29,21 +30,24 @@ public class ConfigService extends PlatformService {
 
   @Override
   protected void doInit() {
-    Config config = getAppConfig();
-    serviceName = config.getString(SERVICE_NAME_CONFIG);
-    serverPort = config.getInt(SERVICE_PORT_CONFIG);
-    LOG.info("Creating {} on port {}", serviceName, serverPort);
+    serverPort = getAppConfig().getInt(SERVICE_PORT_CONFIG);
+    LOG.info("Creating {} on port {}", getServiceName(), serverPort);
 
-    configStore = getConfigStore(config);
-    ConfigServiceGrpcImpl configServiceGrpcImpl = new ConfigServiceGrpcImpl(configStore);
-    configServer = ServerBuilder.forPort(serverPort)
-        .addService(InterceptorUtil.wrapInterceptors(configServiceGrpcImpl))
-        .build();
+    configStore = getConfigStore();
+    ManagedChannel configChannel =
+        ManagedChannelBuilder.forAddress("localhost", serverPort).usePlaintext().build();
+    this.getLifecycle().shutdownComplete().thenRun(configChannel::shutdown);
+
+    configServer =
+        ServerBuilder.forPort(serverPort)
+            .addService(wrapInterceptors(new ConfigServiceGrpcImpl(configStore)))
+            .addService(wrapInterceptors(new SpacesConfigServiceImpl(configChannel)))
+            .build();
   }
 
   @Override
   protected void doStart() {
-    LOG.info("Attempting to start {} on port {}", serviceName, serverPort);
+    LOG.info("Attempting to start {} on port {}", getServiceName(), serverPort);
     try {
       configServer.start();
       LOG.info("Started Config Service on port {}", serverPort);
@@ -62,7 +66,7 @@ public class ConfigService extends PlatformService {
 
   @Override
   protected void doStop() {
-    LOG.info("Shutting down service: {}", serviceName);
+    LOG.info("Shutting down service: {}", getServiceName());
     while (!configServer.isShutdown()) {
       configServer.shutdown();
       try {
@@ -79,15 +83,10 @@ public class ConfigService extends PlatformService {
     return configStore.healthCheck();
   }
 
-  @Override
-  public String getServiceName() {
-    return serviceName;
-  }
-
-  private ConfigStore getConfigStore(Config config) {
+  private ConfigStore getConfigStore() {
     try {
       ConfigStore configStore = new DocumentConfigStore();
-      configStore.init(config);
+      configStore.init(getAppConfig().getConfig(GENERIC_CONFIG_SERVICE_CONFIG));
       return configStore;
     } catch (Exception e) {
       throw new RuntimeException("Error in getting or initializing config store", e);
--- a/spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
+++ b/spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
@@ -1,0 +1,18 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+
+package org.hypertrace.spaces.config.service.v1;
+
+
+message SpaceConfigRule {
+  string id = 1;
+  oneof rule_data {
+    AttributeValueData attribute_value_data = 2;
+  }
+}
+
+message AttributeValueData {
+  string attribute_scope = 1;
+  string attribute_key = 2;
+}
--- a/spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
+++ b/spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
@@ -4,15 +4,14 @@ option java_multiple_files = true;
 
 package org.hypertrace.spaces.config.service.v1;
 
-
 message SpaceConfigRule {
   string id = 1;
   oneof rule_data {
-    AttributeValueData attribute_value_data = 2;
+    AttributeValueRuleData attribute_value_rule_data = 2;
   }
 }
 
-message AttributeValueData {
+message AttributeValueRuleData {
   string attribute_scope = 1;
   string attribute_key = 2;
 }
--- a/spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/spaces_config_service.proto
+++ b/spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/spaces_config_service.proto
@@ -20,7 +20,7 @@ service SpacesConfigService {
 
 message CreateRuleRequest {
   oneof rule_data {
-    AttributeValueData attribute_value_data = 1;
+    AttributeValueRuleData attribute_value_rule_data = 1;
   }
 }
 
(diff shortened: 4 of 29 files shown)


spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/spaces_config_service.proto
```diff
@@ -0,0 +1,52 @@
...
+  oneof rule_data {
+    AttributeValueData attribute_value_data = 1;
+  }
+}
+
+message CreateRuleResponse {
+  SpaceConfigRule rule = 1;
+}
+
+message GetRulesRequest {
```
pavan-traceable, line 31: I don't have much context here, but what if there are so many rules, you want to get either few of them or only one rule

config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
```diff
@@ -0,0 +1,24 @@
+package org.hypertrace.config.proto.converter;
+
+import com.google.protobuf.InvalidProtocolBufferException;
+import com.google.protobuf.Message;
+import com.google.protobuf.MessageOrBuilder;
+import com.google.protobuf.Value;
+import com.google.protobuf.util.JsonFormat;
+
+public class ConfigProtoConverter {
```
pavan-traceable, line 9: This class converts from Proto to Json, should we add that to the class name itself ?

spaces-config-service-impl/src/test/java/org/hypertrace/space/config/service/SpaceConfigRuleConverterImplTest.java
```diff
@@ -0,0 +1,75 @@
...
+  private static final SpaceConfigRule RULE =
+      SpaceConfigRule.newBuilder()
+          .setId("test-id")
+          .setAttributeValueData(
+              AttributeValueData.newBuilder()
+                  .setAttributeKey("attrKey")
+                  .setAttributeScope("attrScope"))
+          .build();
+
+  private static final Value VALUE =
```
pavan-traceable, line 28: This is defined in 3 places. Can we share this ?

config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
GurtejSohi: Conversion from proto to json is just an intermediate step.
The 2 methods in the class are used to convert the `Value` proto object to another proto object(via `Message.Builder`) and vice-versa.

config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
```diff
@@ -29,21 +30,24 @@ public ConfigService(ConfigClient configClient) {
...
+    LOG.info("Creating {} on port {}", getServiceName(), serverPort);
 
-    configStore = getConfigStore(config);
-    ConfigServiceGrpcImpl configServiceGrpcImpl = new ConfigServiceGrpcImpl(configStore);
-    configServer = ServerBuilder.forPort(serverPort)
-        .addService(InterceptorUtil.wrapInterceptors(configServiceGrpcImpl))
-        .build();
+    configStore = getConfigStore();
+    ManagedChannel configChannel =
+        ManagedChannelBuilder.forAddress("localhost", serverPort).usePlaintext().build();
```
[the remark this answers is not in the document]
GurtejSohi: We can switch to the in-process transport (https://grpc.github.io/grpc-java/javadoc/io/grpc/inprocess/package-summary.html). Either I can make the changes in a separate PR or if you want, you can make them in this PR only.

config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
```diff
@@ -0,0 +1,24 @@
...
+  private static final JsonFormat.Printer JSON_PRINTER = JsonFormat.printer();
+  private static final JsonFormat.Parser JSON_PARSER = JsonFormat.parser().ignoringUnknownFields();
+
+  public static void mergeFromValue(Value value, Message.Builder builder)
+      throws InvalidProtocolBufferException {
+    JSON_PARSER.merge(JSON_PRINTER.print(value), builder);
```
GurtejSohi, line 15: I think this would throw an error in the case where `value` indicates an empty/null value (`Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build()`). Need to take care of that.

Might as well be helpful if we have a test for this.

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+
+package org.hypertrace.spaces.config.service.v1;
+
+
```
GurtejSohi, line 7: nit : extra line

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/spaces_config_service.proto
```diff
@@ -0,0 +1,52 @@
...
+message GetRulesResponse {
+  repeated SpaceConfigRule rules = 1;
+}
+
+message UpdateRuleRequest {
+  SpaceConfigRule updated_rule = 1;
+}
+
+message UpdateRuleResponse {
+  SpaceConfigRule rule = 1;
```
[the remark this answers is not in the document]
GurtejSohi: I think we'll emit an error instead of this response if update is not successful. That eliminates the need of a boolean success field.

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
```diff
@@ -0,0 +1,18 @@
...
+message SpaceConfigRule {
+  string id = 1;
+  oneof rule_data {
+    AttributeValueData attribute_value_data = 2;
+  }
+}
+
+message AttributeValueData {
```
GurtejSohi, line 15: Can we just call it `AttributeData` as it doesn't have any attribute value?

config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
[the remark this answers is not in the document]
aaron-steinfeld: Let's do it separately

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/spaces_config_service.proto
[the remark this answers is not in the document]
aaron-steinfeld: Correct - errors don't result in a successful response.

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/spaces_config_service.proto
aaron-steinfeld: We don't currently have a use case where we don't need all rules. Also, for now the expected number of rules is between 0 and 1 - as the feature is built out further for more use cases, we will revisit this.

The two current use cases are
- the space enrichment evaluation, in which case all rules must be evaluated
- The space configuration, in which case all rules are displayed in a table (currently no paging UX for that)

config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
aaron-steinfeld: Right - @GurtejSohi correctly summarizes this. Given that, do we have any name suggestions? I thought of naming it `ConfigProtoValueConverter` but felt that didn't lend any more clarity.

spaces-config-service-impl/src/test/java/org/hypertrace/space/config/service/SpaceConfigRuleConverterImplTest.java
aaron-steinfeld: This is test data - I'd rather not share it between tests to keep things explicit and tests decoupled.

config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
aaron-steinfeld: I think this is debatable when looking at the generic case which this converter captures. If my schema'd message has a valid config using only its default values, such behavior would break those conversions. 

For example, imagine my message had a single boolean field, enabled/disabled:
```
message MyConfig {
    bool enabled = 1;
```

The disabled case would serialize as an empty message due to how proto handles default values, and it would deserialize back to us as an empty value. Currently, this converter would then correctly return an instance of `MyConfig` which would signify disabled, but the proposed change would cause it to error.

I'd prefer to leave validation of the resulting message to the caller.

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
aaron-steinfeld: Fixed

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
aaron-steinfeld: Went back and forth on this - the rule type, in plain english, means "generate a space for a span based on the described attribute's value" - so you're right it doesn't have an attribute value, but since this is the configuration for a rule it's describing how to get one. What about `AttributeValueRuleData`?

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/spaces_config_service.proto
```diff
@@ -0,0 +1,52 @@
...
+}
+
+message DeleteRuleRequest {
+  string id = 1;
+}
+
+message DeleteRuleResponse {
+}
```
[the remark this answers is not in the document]
aaron-steinfeld: Responded above

pavan-traceable (approved):
Changes LGTM.

config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
GurtejSohi: Actually, empty and null value are different here(I confused the two in my previous comment. Sorry for that). I agree with you in the case of empty value.

But now consider this case of null value - we make a `getConfig` call to the generic config service for a config resource for which no config is defined yet. In such case, generic config service returns a null value (actually a representation for null) - `Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build()`. Now if we pass this value to this method with the appropriate builder, it will throw an error as the `JSON_PRINTER` will print `null` and `JSON_PARSER` will be unable to parse it.

spaces-config-service-api/src/main/proto/org/hypertrace/spaces/config/service/v1/space_config_rule.proto
GurtejSohi: This sounds better

config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
aaron-steinfeld: Gotcha - you're talking about the case where the value returned is a top level null. 

First, what are the scenarios where this is returned? Any value that had been written would not be a top level null. Is it only when I ask for a config by context(s) and it doesn't exist (nor a default)? In that case, should the service itself throw an error or return a null? I could be convinced either way there, but I'd probably lean towards throwing as that's generally what API semantics indicate (you'd throw a 404 in the case of REST).

Second, assuming we are returning null in that case, what would we expect this converter do? The caller is expecting a message back in their schema - so the only options I can think of (feel free to add alternate ideas) are:
- Throw: this is what it would currently do as you mentioned `Unexpected exception thrown: java.lang.RuntimeException: com.google.protobuf.InvalidProtocolBufferException: Expect message object but got: null`
- Optional: I think returning an optional for some forms of invalid data but throwing for others is confusing. so this means we'd have to swallow and log any conversion errors and return an empty optional. The caller would likely end up just throwing again anyway (since this is presumably a get call that can't return), this time without the context of the error.
- DefaultInstance: This feels wrong for the same reasons discussed above around empty message - an empty message may be valid and now be indistinguishable from an invalid null.

GurtejSohi: > Is it only when I ask for a config by context(s) and it doesn't exist (nor a default)? 

Yes, that's the only case. 

> In that case, should the service itself throw an error or return a null?

I think this depends on the service. As you pointed, the service might want to throw an error. But this may not be the case always. For instance, the service may want to get config for a config resource, do something with it based on the request(it may want to have its own merging logic) and then upsert it. So, the first time it tries to get a config for a config resource, it will get a null value.

> Second, assuming we are returning null in that case, what would we expect this converter do?

I think returning an `Optional` would be fine. We just check if `value` is null(i.e. `value.equals(Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build())`), then return an empty optional. We don't have to swallow any conversion errors here, right?

aaron-steinfeld: > I think this depends on the service. As you pointed, the service might want to throw an error. But this may not be the case always. For instance, the service may want to get config for a config resource, do something with it based on the request(it may want to have its own merging logic) and then upsert it. So, the first time it tries to get a config for a config resource, it will get a null value.

It's certainly true that we can make either work. The service can throw, the caller can catch it, recover and proceed - or the service can return null, the caller can check it and throw. So it's more about the api design rather than what use cases does each enable. My preference is that the api exposes one error path rather than two separate ones.

> I think returning an Optional would be fine. We just check if value is null(i.e. value.equals(Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build())), then return an empty optional. We don't have to swallow any conversion errors here, right?

I think this directly follows from the prior point. If we maintain the response/null/error behavior of the service, the optional would allow us to mimic it and probably makes sense. If we think that behavior should coalesce to just response/error, this should too.

GurtejSohi (approved):

config-proto-converter/src/main/java/org/hypertrace/config/proto/converter/ConfigProtoConverter.java
aaron-steinfeld: @GurtejSohi - I do think the converter should match the behavior of the api, but right now they're inconsistent. Do we think it makes sense to make the behavior change and throw instead of returning null (in which case, no change required here) or to leave the behavior as is and update the converter?

GurtejSohi: > I do think the converter should match the behavior of the api, but right now they're inconsistent.

Oh, right. For a second, I thought the generic config service was in another repo :)

> Do we think it makes sense to make the behavior change and throw instead of returning null (in which case, no change required here) or to leave the behavior as is and update the converter?

My preference is that we maintain response/null/error behaviour (and update the converter) because I think it helps the consumer to distinguish between "config not found" and "server error" without having to check/process the error message.
But, I'm fine with the other approach as well. Please feel free to take the call.

aaron-steinfeld: I'm going to take it in a separate PR, but unless anyone has strong feelings (@buchi-busireddy  ?), I'm going to change the behavior of the API to return a status `5` https://grpc.github.io/grpc/core/md_doc_statuscodes.html (grpc equivalent to 404) if the response is not found. My reasoning is that, in the not found case we're going to have to parse something - either the error or the response, to disambiguate it from what we're calling a "real" error or "valid" response. I'd rather use the GRPC standard error code for this case rather than building an api-specific body that indicates the same. The client utility I see as completely orthogonal.

PR #10, opened 2021-01-08
chore: fix snyk update and fix buf

aaron-steinfeld:
Updating snyk to fix vulnerabilities after registering projects. Also updated buf and fixed url format.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -4,21 +4,21 @@ on:
     branches:
       - main
   pull_request_target:
-    branches: 
+    branches:
       - main
 
 jobs:
   build:
     runs-on: ubuntu-20.04
     steps:
-       # Set fetch-depth: 0 to fetch commit history and tags for use in version calculation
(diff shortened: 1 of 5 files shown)


aaron-steinfeld:
Validate protos is failing because it's scoped to pull_request_target which won't pick up config changes until merged. I verified the fix here: https://github.com/hypertrace/config-service/runs/1670668738

JBAhire (approved):
We can move validate protos to pr-test workflow if we have ghcr image of buf or if we can manually install it. 

Considering there will be limited changes in this repo, I won't even worry about rate-limiting now and then build and push buf image to ghcr afterwards.

aaron-steinfeld:
> We can move validate protos to pr-test workflow if we have ghcr image of buf or if we can manually install it. 
> 
> Considering there will be limited changes in this repo, I won't even worry about rate-limiting now and then build and push buf image to ghcr afterwards. 

It's easy enough to download and use it - the container just used to be easier. We can look at switching to a buf setup style action if we run into issues.

GurtejSohi (approved):

PR #17, opened 2021-02-01
fixing an initialisation issue when using postgres as docstore

kotharironak:
The listCollection api returns the list of tables prefix with db name.
```
INFO  o.h.c.s.s.DocumentConfigStore - list of collections:[default_db.attribute_metadata, default_db.config_bootstrap_status, default_db.configurations, default_db.entity_types, default_db.entity_type_relations]
```
So, the method `getOrCreateCollection` was failing on postgres collection side if a table already exists.

```
postgres                | 2021-02-01 08:16:20.136 UTC [76] ERROR:  relation "configurations" already exists
```

Currently, there is an assumption on dbname (`default_db`) while using mongo as docstore, we will have to fix that first. For now, 
We don't have to explicitly check if the collection exists while using `mongo` as docstore. On postgres side, it has been taken care of inside getCollection api.

### DIFF
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
@@ -55,7 +55,7 @@ public class DocumentConfigStore implements ConfigStore {
   @Override
   public void init(Config config) {
     datastore = initDataStore(config);
-    this.collection = getOrCreateCollection(datastore);
+    this.collection = this.datastore.getCollection(CONFIGURATIONS_COLLECTION);;
   }
 
   private Datastore initDataStore(Config config) {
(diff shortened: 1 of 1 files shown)


config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -55,7 +55,7 @@
   @Override
   public void init(Config config) {
     datastore = initDataStore(config);
-    this.collection = getOrCreateCollection(datastore);
+    this.collection = this.datastore.getCollection(CONFIGURATIONS_COLLECTION);;
```
pavan-traceable, line 58: Won't this affect other stores ?

kotharironak: Added the context and description. 
Postgres handles currently in getCollection API
- https://github.com/hypertrace/document-store/blob/main/document-store/src/main/java/org/hypertrace/core/documentstore/postgres/PostgresDatastore.java#L116

kotharironak: Have tested locally running Hypetrace with postgres
Screenshot 2021-02-01 at 5 38 57 PM
Screenshot 2021-02-01 at 5 40 17 PM

GurtejSohi:
> We don't have to explicitly check if the collection exists while using mongo as docstore.

What happens if the collection does not exist? Does the method `getCollection()` create it in that case?

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
kotharironak: We had a restructuring ticket for the same, but it was not prioritized so far - https://github.com/hypertrace/hypertrace/issues/123

kotharironak:
> > We don't have to explicitly check if the collection exists while using mongo as docstore.
> 
> What happens if the collection does not exist? Does the method `getCollection()` create it in that case?

Yes, mongo takes care of it whenever any document is added. This is how it is used currently in other services - https://github.com/hypertrace/entity-service/blob/main/entity-service-impl/src/main/java/org/hypertrace/entity/data/service/EntityDataServiceImpl.java#L68 - while using mongo and so, Postgres side, we had to create a collection inside `getCollection` api implementation.

GurtejSohi (approved):

PR #19, opened 2021-02-04
Upsert overriding with the new config to be upserted

anujgoyal1:
This change was needed to support the update of the fields with default proto value. We can't distinguish if the default value was being set(updated) or it's just the default proto value.
eg: If a flag  in config is marked as true, now we want to update it to false, it's not possible with the current merging/partial update logic. So, for now, support is for full update of the config.


aaron-steinfeld (commented):
Please update `config_service.proto` docs to indicate new behavior as well as updating the description to describe the motivation for this change.

GurtejSohi (approved):

PR #27, opened 2021-02-23
refactor: move around test-fixtures for publishing

aaron-steinfeld:


spaces-config-service-impl/build.gradle.kts
```diff
@@ -18,7 +18,7 @@ dependencies {
   testImplementation("org.junit.jupiter:junit-jupiter:5.7.0")
   testImplementation("org.mockito:mockito-core:3.7.0")
   testImplementation("org.mockito:mockito-junit-jupiter:3.7.0")
-  testImplementation(testFixtures(project(":config-service")))
+  testImplementation(testFixtures(project(":config-service-api")))
```
pavan-traceable, line 21: testFixtures on api doesn't look convincing. Should we add new module like utils and move these ?

aaron-steinfeld: Why? Isn't this exactly what the test fixtures are associated with? They're a way to write tests against the api without needing the real implementation, and are used by consumers that are depending on the api.

pavan-traceable: Thanks, i am not aware of this.

Issue #25, opened 2021-02-22
Expose creation and update timestamp in get config APIs


GurtejSohi:
## Use Case
The client may be interested in the creation and/or update timestamp for config that it is dealing with. For example, client may want to sort all the configs based on the creation timestamp.

For this, we need to expose the creation and update timestamp in `GetConfig` and `GetAllConfigs` API.

## Proposal
We can add 2 extra fields in `GetConfigResponse` which takes care of the `GetConfig` API:
```
message GetConfigResponse {
  google.protobuf.Value config = 1;
  int64 creation_timestamp = 2; // new field
  int64 update_timestamp = 3;  // new field
}
```

We can add 2 extra fields in `ContextSpecificConfig` which takes care of the `GetAllConfigs` API:
```
message GetAllConfigsResponse {
  repeated ContextSpecificConfig context_specific_configs = 1;
}

message ContextSpecificConfig {
  string context = 1;
  google.protobuf.Value config = 2;
  int64 creation_timestamp = 3; // new field
  int64 update_timestamp = 4;  // new field
}
```

From implementation perspective, we'll now have to call `getConfig` while upserting any config so as to get the `creationTimestamp` for that config.

aaron-steinfeld:
Seems right to me. We could add them on the upsert response too, but I'm fine ignoring that until there's a use case.

PR #28, opened 2021-02-26
Expose creation and update timestamp in getAllConfigs API

GurtejSohi:
## Description
Addresses https://github.com/hypertrace/config-service/issues/25

### Testing
Unit tests and integration tests

### Checklist:

### DIFF
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -68,7 +68,7 @@ message GetAllConfigsRequest {
 }
 
 message GetAllConfigsResponse {
-  // list of config values along with the associated contexts
+  // list of config values along with the associated contexts, sorted in the ascending order of their creation time
   repeated ContextSpecificConfig context_specific_configs = 1;
 }
 
@@ -78,6 +78,12 @@ message ContextSpecificConfig {
 
   // config value which applies to the above context
   google.protobuf.Value config = 2;
+
+  // timestamp (in milliseconds since epoch) at which this config was created
+  int64 creation_timestamp = 3;
+
+  // timestamp (in milliseconds since epoch) at which this config was last updated
+  int64 update_timestamp = 4;
 }
 
 message DeleteConfigRequest {
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
@@ -38,6 +38,8 @@ public class ConfigDocument implements Document {
   public static final String VERSION_FIELD_NAME = "configVersion";
   public static final String USER_ID_FIELD_NAME = "userId";
   public static final String CONFIG_FIELD_NAME = "config";
+  public static final String CREATION_TIMESTAMP_FIELD_NAME = "creationTimestamp";
+  public static final String UPDATE_TIMESTAMP_FIELD_NAME = "updateTimestamp";
 
   @JsonProperty(value = RESOURCE_FIELD_NAME)
   String resourceName;
@@ -62,6 +64,12 @@ public class ConfigDocument implements Document {
   @JsonProperty(value = CONFIG_FIELD_NAME)
   Value config;
 
+  @JsonProperty(value = CREATION_TIMESTAMP_FIELD_NAME)
+  long creationTimestamp;
+
+  @JsonProperty(value = UPDATE_TIMESTAMP_FIELD_NAME)
+  long updateTimestamp;
+
   @JsonCreator(mode = Mode.PROPERTIES)
   public ConfigDocument(@JsonProperty(RESOURCE_FIELD_NAME) String resourceName,
       @JsonProperty(RESOURCE_NAMESPACE_FIELD_NAME) String resourceNamespace,
@@ -69,7 +77,9 @@ public class ConfigDocument implements Document {
       @JsonProperty(CONTEXT_FIELD_NAME) String context,
       @JsonProperty(VERSION_FIELD_NAME) long configVersion,
       @JsonProperty(USER_ID_FIELD_NAME) String userId,
-      @JsonProperty(CONFIG_FIELD_NAME) Value config) {
+      @JsonProperty(CONFIG_FIELD_NAME) Value config,
+      @JsonProperty(CREATION_TIMESTAMP_FIELD_NAME) long creationTimestamp,
+      @JsonProperty(UPDATE_TIMESTAMP_FIELD_NAME) long updateTimestamp) {
     this.resourceName = resourceName;
     this.resourceNamespace = resourceNamespace;
     this.tenantId = tenantId;
@@ -77,6 +87,8 @@ public class ConfigDocument implements Document {
     this.configVersion = configVersion;
     this.userId = userId;
     this.config = config;
+    this.creationTimestamp = creationTimestamp;
+    this.updateTimestamp = updateTimestamp;
   }
 
   public static ConfigDocument fromJson(String json) throws IOException {
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
@@ -6,6 +6,8 @@ import com.google.common.cache.LoadingCache;
 import com.google.protobuf.Value;
 import com.typesafe.config.Config;
 import java.util.ArrayList;
+import java.util.Collections;
+import java.util.Comparator;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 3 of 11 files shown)


config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -68,7 +68,7 @@ message GetAllConfigsRequest {
 }
 
 message GetAllConfigsResponse {
-  // list of config values along with the associated contexts
+  // list of config values along with the associated contexts, sorted in the ascending order of their creation time
```
GurtejSohi, line 71: I think usually the client would want the configs to be returned in the order of creation time. So, added this extra condition.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -78,6 +78,12 @@ message ContextSpecificConfig {
 
   // config value which applies to the above context
   google.protobuf.Value config = 2;
+
+  // timestamp (in milliseconds since epoch) at which this config was created
+  int64 creation_timestamp = 3;
+
+  // timestamp (in milliseconds since epoch) at which this config was last updated
+  int64 update_timestamp = 4;
```
GurtejSohi, line 86: Haven't added these fields to `getConfig` API because that API can merge the configs belonging to different contexts, each of which will have different creation and update timestamps.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
aaron-steinfeld: That makes sense. If we need to add sort support later, we can

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -72,27 +73,46 @@ public long writeConfig(ConfigResource configResource, String userId, Value conf
...
+        .and(Filter.eq(VERSION_FIELD_NAME, configVersion));
     Query query = new Query();
     query.setFilter(filter);
     Iterator<Document> documentIterator = collection.search(query);
     if (documentIterator.hasNext()) {
       String documentString = documentIterator.next().toJson();
       ConfigDocument configDocument = ConfigDocument.fromJson(documentString);
-      config = configDocument.getConfig();
+      if (!ConfigServiceUtils.isNull(configDocument.getConfig())) {
+        config = ContextSpecificConfig.newBuilder().setConfig(configDocument.getConfig())
```
aaron-steinfeld, line 111: We could return directly here and return an empty config directly on 117, right? No real need to init at 101.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -116,9 +136,13 @@ public Value getConfig(ConfigResource configResource) throws IOException {
       Value config = configDocument.getConfig();
       if (seenContexts.add(context) && !ConfigServiceUtils.isNull(config)) {
         contextSpecificConfigList
-            .add(ContextSpecificConfig.newBuilder().setContext(context).setConfig(config).build());
+            .add(ContextSpecificConfig.newBuilder().setContext(context).setConfig(config)
```
aaron-steinfeld, line 139: could probably pull the `ConfigDocument -> ContextSpecificConfig` into its own function in use in the two spots

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
```diff
@@ -62,21 +64,31 @@
   @JsonProperty(value = CONFIG_FIELD_NAME)
   Value config;
 
+  @JsonProperty(value = CREATION_TIMESTAMP_FIELD_NAME)
+  long creationTimestamp;
```
aaron-steinfeld, line 68: Will we have any backwards compatibility issues? For existing documents, will they just default to have a 0 creation/update timestamp, or would we run into errors?

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
aaron-steinfeld: That makes sense. Eventually, we may have to expose an API that doesn't merge and does expose those (wish getConfig had been named getMergedConfig or something...), but no need until we have a use case. The same argument could actually be made for this API since the main use case as I understand it is internal - the sorted results, but I'm fine either way since it's done.

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
```diff
@@ -63,8 +66,10 @@ static void init() {
   @Test
   void writeConfig() throws IOException {
     List<Document> documentList = Collections
-        .singletonList(getConfigDocument(DEFAULT_CONTEXT, CONFIG_VERSION, config1));
-    when(collection.search(any(Query.class))).thenReturn(documentList.iterator());
+        .singletonList(
+            getConfigDocument(DEFAULT_CONTEXT, CONFIG_VERSION, config1, TIMESTAMP1, TIMESTAMP2));
+    when(collection.search(any(Query.class)))
+        .thenReturn(documentList.iterator(), documentList.iterator());
```
aaron-steinfeld, line 72: Why the second return value?

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
```diff
@@ -76,42 +81,51 @@ void writeConfig() throws IOException {
     Document document = documentCaptor.getValue();
     long newVersion = CONFIG_VERSION + 1;
     assertEquals(new ConfigDocumentKey(configResource, newVersion), key);
-    assertEquals(getConfigDocument(DEFAULT_CONTEXT, newVersion, config1), document);
+    assertEquals(getConfigDocument(DEFAULT_CONTEXT, newVersion, config1, TIMESTAMP1,
+        ((ConfigDocument) document).getUpdateTimestamp()), document);
   }
 
   @Test
   void getConfig() throws IOException {
```
aaron-steinfeld, line 89: Can we add a backwards compatibility test in here? That is, a document without a create/update timestamp

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
GurtejSohi: updated it

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
GurtejSohi: updated it

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
GurtejSohi: Existing documents will default to have the creation and update timestamp as 0. So, we won't run into errors - it's just that it won't have their actual creation timestamp.

Also added a test for `ConfigDocument` to verify this behaviour^.

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
GurtejSohi: `writeConfig` now makes 2 calls to `collection.search()` - first one to get the latest version for this resource and the second one to get the existing config for this resource(so that it can get the creation timestamp).

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
GurtejSohi: Added a backwards compatibility test for `ConfigDocument`. Here, I think it won't make much sense because we would simply be returning a mocked document with creation and update timestamp as 0 in the mocked call.

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
aaron-steinfeld: But given that it's the same value (unless it got changed somewhere else - may have missed that), `thenReturn` supports multiple calls - extra args are if you want to return a different result on the second call than the first.

GurtejSohi: It returns an iterator - so the position of the first iterator would have changed after it is consumed once.
So, I want to return a fresh iterator each time the mocked method is called.

aaron-steinfeld: Ah, gotcha.

aaron-steinfeld (approved):

PR #29, opened 2021-03-03
Expose creation and update timestamp in UpsertConfigResponse

GurtejSohi:
## Description
Extension of https://github.com/hypertrace/config-service/issues/25

### Testing
Updated unit tests

### Checklist:

### DIFF
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -39,6 +39,12 @@ message UpsertConfigRequest {
 message UpsertConfigResponse {
   // config value upserted into the store
   google.protobuf.Value config = 1;
+
+  // timestamp (in milliseconds since epoch) at which this config was created
+  int64 creation_timestamp = 2;
+
+  // timestamp (in milliseconds since epoch) at which this config was last updated
+  int64 update_timestamp = 3;
 }
 
 message GetConfigRequest {
--- a/config-service-impl/build.gradle.kts
+++ b/config-service-impl/build.gradle.kts
@@ -22,6 +22,12 @@ dependencies {
   testImplementation("org.junit.jupiter:junit-jupiter:5.5.2")
   testImplementation("org.mockito:mockito-core:3.3.3")
   testImplementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.3.3")
+
+  constraints {
+    runtimeOnly("org.mongodb:mongodb-driver-sync:4.1.2") {
+      because("https://snyk.io/vuln/SNYK-JAVA-ORGMONGODB-1079241")
+    }
+  }
 }
 
 tasks.test {
--- a/config-service-impl/build.gradle.kts
+++ b/config-service-impl/build.gradle.kts
@@ -13,7 +13,7 @@ dependencies {
   implementation("com.typesafe:config:1.4.0")
   implementation("org.slf4j:slf4j-api:1.7.30")
 
-  implementation("org.hypertrace.core.documentstore:document-store:0.5.3")
+  implementation("org.hypertrace.core.documentstore:document-store:0.5.4")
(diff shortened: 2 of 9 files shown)


aaron-steinfeld (commented):
the mock generic config service doesn't implement the timestamp portions of the response - I think that's fine until needed, but probably worth either documenting in the meantime.

aaron-steinfeld (dismissed):

aaron-steinfeld:
Can you update the vulnerabilities too - either in a separate PR or this one is fine.

GurtejSohi:
> the mock generic config service doesn't implement the timestamp portions of the response - I think that's fine until needed, but probably worth either documenting in the meantime.

Since it was a small change, I've updated the mock generic config service to implement the timestamp portions of the response.

> Can you update the vulnerabilities too - either in a separate PR or this one is fine.

Upgraded the versions of vulnerable libs as pointed out by snyk.

config-service/build.gradle.kts
```diff
@@ -80,6 +80,9 @@ dependencies {
     runtimeOnly("io.netty:netty-handler-proxy:4.1.59.Final") {
       because("https://snyk.io/vuln/SNYK-JAVA-IONETTY-1070799")
     }
+    runtimeOnly("org.eclipse.jetty:jetty-server:9.4.37.v20210219") {
+      because("https://snyk.io/vuln/SNYK-JAVA-ORGECLIPSEJETTY-1080611")
```
aaron-steinfeld, line 84: jetty is from platform-service-framework - upgrading that to latest would be more appropriate than constraining a version here - we only want to use constraints when the vulnerability is coming via a third party we can't update.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -39,6 +39,12 @@ message UpsertConfigRequest {
 message UpsertConfigResponse {
   // config value upserted into the store
   google.protobuf.Value config = 1;
+
+  // timestamp (in milliseconds since epoch) at which this config was created
+  int64 creation_timestamp = 2;
```
jcchavezs, line 44: When it comes to code it feel ambiguous to not to know the unit. How about creation_timestamp_ms?

config-service-impl/build.gradle.kts
```diff
@@ -22,6 +22,12 @@ dependencies {
   testImplementation("org.junit.jupiter:junit-jupiter:5.5.2")
   testImplementation("org.mockito:mockito-core:3.3.3")
   testImplementation("org.hypertrace.core.grpcutils:grpc-client-utils:0.3.3")
+
+  constraints {
+    runtimeOnly("org.mongodb:mongodb-driver-sync:4.1.2") {
```
aaron-steinfeld, line 27: Likewise, this comes from document-store - I see it came up on https://github.com/hypertrace/document-store/pull/37 @kotharironak  - any reason it wasn't updated? Can we get a PR in to fix it since it's leaking out to all consumers

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
aaron-steinfeld: I completely agree, however here I'd prefer consistency, since the same field is already present in other responses in the same service API. For what it's worth, I believe across the entire hypertrace platform timestamps stored as long are always millis (and should have been named that way in the first place) - we don't use seconds or micro/nano anywhere that I'm aware of.

config-service/build.gradle.kts
GurtejSohi: Makes sense. Upgraded platform-service-framework version and removed this constraint

config-service-impl/build.gradle.kts
aaron-steinfeld: Will raise this pr.

aaron-steinfeld (approved):

PR #34, opened 2021-03-05
Fixes #16 : Handle null values gracefully

### LINKED ISSUE #16: Fix the inconsistent error log
```
if (seenContexts.add(context) && !ConfigServiceUtils.isNull(config)) {
        contextSpecificConfigList
            .add(ContextSpecificConfig.newBuilder().setContext(context).setConfig(config).build());
}
```
in `getAllConfig` method of `DocumentConfigStore`, if value of config is null, an error is thrown.
But `deleteConfig` itself sets the value config in of deletedVersion to null. 
So if deleteConfig is called followed by getAllConfigs, it's end up throwing error.

GurtejSohi:
## Description
Fixes https://github.com/hypertrace/config-service/issues/16

Serialization of `NULL_VALUE` (`Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build()`) yields the string - "null", which when deserialized was previously returning the Value as `null`. Although, we handle `null` also using the [isNull() method](https://github.com/hypertrace/config-service/blob/main/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceUtils.java#L73) (which is why it wasn't affecting the client), we also log an error saying that it is unexpected because ideally, we shouldn't be working with `null` - instead we should be using the null placeholder - `NULL_VALUE`.

So, this change tells the `JsonDeserializer` to use the null placeholder (`NULL_VALUE`) on encountering the string - "null", which ensures that we never get `null` config on deserialization.

### Testing
Added unit test

### Checklist:
- [x] I have added tests that prove my fix is effective

### DIFF
--- a/config-service-impl/src/test/java/org/hypertrace/config/service/store/ConfigDocumentTest.java
+++ b/config-service-impl/src/test/java/org/hypertrace/config/service/store/ConfigDocumentTest.java
@@ -8,6 +8,9 @@ import static org.hypertrace.config.service.TestUtils.getConfig1;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
 import com.google.common.io.Resources;
+import com.google.protobuf.NullValue;
+import com.google.protobuf.Value;
+import com.google.protobuf.util.JsonFormat;
 import java.io.IOException;
 import java.nio.charset.Charset;
 import org.junit.jupiter.api.DisplayName;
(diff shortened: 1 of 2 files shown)


config-service-impl/src/test/java/org/hypertrace/config/service/store/ConfigDocumentTest.java
```diff
@@ -23,6 +26,15 @@ void convertToAndFromJson() throws IOException {
     assertEquals(configDocument, ConfigDocument.fromJson(configDocument.toJson()));
   }
 
+  @Test
+  void convertDocumentContainingNullValue() throws IOException {
+    long timestamp = System.currentTimeMillis();
+    Value nullValue = Value.newBuilder().setNullValue(NullValue.NULL_VALUE).build();
+    ConfigDocument configDocument = new ConfigDocument(RESOURCE_NAME, RESOURCE_NAMESPACE,
+        TENANT_ID, DEFAULT_CONTEXT, 15, "user1", nullValue, timestamp, timestamp);
+    assertEquals(configDocument, ConfigDocument.fromJson(configDocument.toJson()));
```
GurtejSohi, line 35: This assertion fails without the above changes in `ConfigDocument` as in that case, the config document - `ConfigDocument.fromJson(configDocument.toJson())` would have config value as `null`.

aaron-steinfeld (approved):

PR #40, opened 2021-04-29
fix: update mock generic config service

aaron-steinfeld:
## Description
When using inline mocks, we want to use the real bind service in our mock config service. Also bumped some services.

### DIFF
--- a/config-service-api/src/testFixtures/java/org/hypertrace/config/service/test/MockGenericConfigService.java
+++ b/config-service-api/src/testFixtures/java/org/hypertrace/config/service/test/MockGenericConfigService.java
@@ -62,6 +62,11 @@ public class MockGenericConfigService {
       Mockito.mock(
           ConfigServiceImplBase.class,
           invocation -> { // Error if unmocked called so we don't hang waiting for streamobserver
+            if (invocation
+                .getMethod()
+                .equals(ConfigServiceImplBase.class.getMethod("bindService"))) {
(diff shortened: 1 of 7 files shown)


config-service-api/src/testFixtures/java/org/hypertrace/config/service/test/MockGenericConfigService.java
```diff
@@ -62,6 +62,11 @@
       Mockito.mock(
           ConfigServiceImplBase.class,
           invocation -> { // Error if unmocked called so we don't hang waiting for streamobserver
+            if (invocation
```
pavan-traceable, line 65: Just curious, why is it needed now ? We have mocked all methods in config service right.

pavan-traceable (approved):

config-service-api/src/testFixtures/java/org/hypertrace/config/service/test/MockGenericConfigService.java
aaron-steinfeld: So I believe what happened is the behavior was dependent on the mock maker. When the constructor is called, it adds the mock service to the server builder, which tries to bind it by calling bindService. Using the default mock maker as we normally do it works (although I'm not quite sure why - it could be a discrepancy on init order?), but using the inline mock maker (for support of final and static mocking) it was failing due to the default invocation behavior here.

PR #41, opened 2021-05-27
feat: return deleted config in delete config response

GurtejSohi:
## Description
We need to return the deleted config value in `DeleteConfigResponse`

### Testing
Updated unit test and integration tests

### Checklist:

### DIFF
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -103,6 +103,7 @@ message DeleteConfigRequest {
   string context = 3;
 }
 
-// empty response
 message DeleteConfigResponse {
+  // deleted config
+  ContextSpecificConfig context_specific_config = 1;
 }
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -104,6 +104,5 @@ message DeleteConfigRequest {
 }
 
 message DeleteConfigResponse {
-  // deleted config
-  ContextSpecificConfig context_specific_config = 1;
+  ContextSpecificConfig deleted_config = 1;
 }
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
@@ -108,6 +108,14 @@ public class ConfigServiceGrpcImpl extends ConfigServiceGrpc.ConfigServiceImplBa
     try {
       ConfigResource configResource = getConfigResource(request);
       ContextSpecificConfig configToDelete = configStore.getConfig(configResource);
+
+      // if configToDelete is null/empty (i.e. config value doesn't exist or is already deleted),
+      // then throw NOT_FOUND exception
+      if (ConfigServiceUtils.isNull(configToDelete.getConfig())) {
+        responseObserver.onError(Status.NOT_FOUND.asException());
+        return;
+      }
+
       // write an empty config for the specified config resource. This maintains the versioning.
       configStore.writeConfig(configResource, getUserId(), emptyValue());
       responseObserver.onNext(
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
@@ -106,9 +106,12 @@ public class ConfigServiceGrpcImpl extends ConfigServiceGrpc.ConfigServiceImplBa
   public void deleteConfig(
       DeleteConfigRequest request, StreamObserver<DeleteConfigResponse> responseObserver) {
     try {
+      ConfigResource configResource = getConfigResource(request);
+      ContextSpecificConfig configToDelete = configStore.getConfig(configResource);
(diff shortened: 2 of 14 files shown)


config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -103,6 +103,7 @@ message DeleteConfigRequest {
   string context = 3;
 }
 
-// empty response
 message DeleteConfigResponse {
+  // deleted config
+  ContextSpecificConfig context_specific_config = 1;
```
aaron-steinfeld, line 108:
```suggestion
  ContextSpecificConfig deleted_config = 1;
```

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -106,9 +106,12 @@ public void getAllConfigs(
   public void deleteConfig(
       DeleteConfigRequest request, StreamObserver<DeleteConfigResponse> responseObserver) {
     try {
+      ConfigResource configResource = getConfigResource(request);
```
aaron-steinfeld, line 109: This will error if the value didn't exist, which is different than the previous behavior, right? That seems more correct to me, just want to point it out.

config-service/src/integrationTest/java/org/hypertrace/config/service/ConfigServiceIntegrationTest.java
```diff
@@ -162,14 +162,20 @@ public void testDeleteConfig() {
     upsertConfig(RESOURCE_NAME, RESOURCE_NAMESPACE, TENANT_1, Optional.of(CONTEXT_1), config2);
 
     // delete config with context
-    deleteConfig(RESOURCE_NAME, RESOURCE_NAMESPACE, TENANT_1, CONTEXT_1);
+    ContextSpecificConfig deletedConfig =
```
aaron-steinfeld, line 165: please also update `MockGenericConfigService`

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: No, why will this line throw an error?

I'm assuming you wanted to highlight the next line - #110 ? (i.e. `configStore.getConfig()` would throw an error if the value didn't exist?) - but that's not right - `configStore.getConfig()` would return empty config value in that case (and so, would this `deleteConfig` API).

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
GurtejSohi: Updated it

config-service/src/integrationTest/java/org/hypertrace/config/service/ConfigServiceIntegrationTest.java
GurtejSohi: Oh, forgot about it initially! Updated it now.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aaron-steinfeld: sorry, didn't dive in - I thought getConfig throws if resource isn't present, but maybe that logic comes in at a higher layer. I'm not sure which is preferable, but it seems wrong to write an empty if no value exists, no?

aaron-steinfeld (approved):

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
GurtejSohi: > I'm not sure which is preferable, but it seems wrong to write an empty if no value exists, no?

We're treating empty config value as equivalent to no config value (in the store). So, going by that, there is no difference if no value exists or empty value exists in the store.

aaron-steinfeld: but wouldn't this do things like increment the version and set a create time, so when we do write a config for the first time, its metadata will be off?

GurtejSohi: Actually, I take that back. Thinking from the API perspective - deleting a non-existing(whether no config value was defined or the config value was deleted previously) config should throw a `Status.NOT_FOUND` exception, right? So, if `configStore.getConfig()`(line #110) returns an empty config, then we should throw that exception, right?

GurtejSohi: Sorry, I just saw your comment after I typed my above comment. 

> but wouldn't this do things like increment the version and set a create time, so when we do write a config for the first time, its metadata will be off?

Agree. I'll make the above change then.

aaron-steinfeld: > deleting a non-existing(whether no config value was defined or the config value was deleted previously) config should throw a Status.NOT_FOUND exception, right?

Yeah that was the other part I was kind of toying with above, but that behavior seems right to me.

GurtejSohi: Made this change. Also added a unit test for deleting a non-existing config.

aaron-steinfeld (approved):

pavan-traceable (approved):

PR #48, opened 2021-08-23
ENG-11061: Adding implementation corresponding to tags config service apis

kamaleshnneerasa:


tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/ConfigServiceCoordinatorImpl.java
```diff
@@ -0,0 +1,113 @@
...
+import org.hypertrace.config.service.v1.GetAllConfigsResponse;
+import org.hypertrace.config.service.v1.GetConfigRequest;
+import org.hypertrace.config.service.v1.GetConfigResponse;
+import org.hypertrace.config.service.v1.UpsertConfigRequest;
+import org.hypertrace.config.service.v1.UpsertConfigResponse;
+import org.hypertrace.core.grpcutils.client.RequestContextClientCallCredsProviderFactory;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.tag.config.service.v1.Tag;
+
+public class ConfigServiceCoordinatorImpl implements ConfigServiceCoordinator {
```
jcchavezs, line 26: Shall this be final? I don't think we need to extend this class.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/ConfigServiceCoordinatorImpl.java
```diff
@@ -0,0 +1,113 @@
...
+            .setResourceNamespace(TAG_RESOURCE_NAMESPACE)
+            .build();
+    GetAllConfigsResponse getAllConfigsResponse =
+        getAllConfigs(requestContext, getAllConfigsRequest);
+    return getAllConfigsResponse.getContextSpecificConfigsList().stream()
+        .map(configResponse -> convertToTagFromGeneric(configResponse.getConfig()))
+        .collect(Collectors.toList());
+  }
+
+  public GetAllConfigsResponse getAllConfigs(RequestContext context, GetAllConfigsRequest request) {
```
jcchavezs, line 83: Does this method need to be public? I don't see anyone using it.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigConstants.java
```diff
@@ -0,0 +1,9 @@
+package org.hypertrace.tag.config.service;
+
+public class TagConfigConstants {
+
+  private TagConfigConstants() {}
+
+  public static final String TAG_CONFIG = "tag-config";
```
jcchavezs, line 7: I am usually curious about the pattern `MY_CONSTANT="my-constant"`, I don't see what is the value added, I think if at least we could specify what is this constant describing would be more helpful e.g. TAG_CONFIG_KEY or CONFIG_KEY (the TAG_ prefix does not seem needed).

tags-config-service-impl/src/test/java/org/hypertrace/tag/config/service/TagConfigServiceImplTest.java
```diff
@@ -0,0 +1,158 @@
...
+import org.hypertrace.tag.config.service.v1.Tag;
+import org.hypertrace.tag.config.service.v1.TagConfigServiceGrpc;
+import org.hypertrace.tag.config.service.v1.TagConfigServiceGrpc.TagConfigServiceBlockingStub;
+import org.hypertrace.tag.config.service.v1.UpdateTagRequest;
+import org.hypertrace.tag.config.service.v1.UpdateTagResponse;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class TagConfigServiceImplTest {
```
jcchavezs, line 28: nit: this could be final

jcchavezs, line 28: I was trying to read the tests, I believe it is a CRUD test but it feels very hard to read. I might feel to a bit of overhead to abstract the steps into private functions but at least some comments would help describing what exactly are we doing and what are we trying to achieve. On the other hand, all this tests look happy paths, do we have any relevant non-happy path to include?

nit: this class could be final.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/ConfigServiceCoordinatorImpl.java
kamaleshnneerasa: Will make it final.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/ConfigServiceCoordinatorImpl.java
kamaleshnneerasa: Yes will make it private.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigConstants.java
kamaleshnneerasa: Changed the names. Hopefully, makes more sense now.

tags-config-service-impl/src/test/java/org/hypertrace/tag/config/service/TagConfigServiceImplTest.java
kamaleshnneerasa: Added few comments and refactored the code for better understanding. There are no non-happy paths.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    List<Tag> tagList = configServiceCoordinator.getAllTags(requestContext);
+    responseObserver.onNext(GetTagsResponse.newBuilder().addAllTags(tagList).build());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void updateTag(
+      UpdateTagRequest request, StreamObserver<UpdateTagResponse> responseObserver) {
+    RequestContext requestContext = RequestContext.CURRENT.get();
+    Tag updateTag = request.getTag();
```
saxenakshitiz, line 64: We should get tag to check if it exists, before update

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
```diff
@@ -0,0 +1,79 @@
...
+    responseObserver.onNext(UpdateTagResponse.newBuilder().setTag(updatedTag).build());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void deleteTag(
+      DeleteTagRequest request, StreamObserver<DeleteTagResponse> responseObserver) {
+    RequestContext requestContext = RequestContext.CURRENT.get();
+    String tagId = request.getId();
+    configServiceCoordinator.deleteTag(requestContext, tagId);
```
saxenakshitiz, line 75: We should get tag to check if it exists, before delete

tags-config-service-impl/src/test/java/org/hypertrace/tag/config/service/TagConfigServiceImplTest.java
```diff
@@ -0,0 +1,132 @@
...
+    assertEquals(Set.copyOf(createdTagsList), Set.copyOf(getTags));
+  }
+
+  @Test
+  void test_updateTag() {
+    List<Tag> createdTagsList = createTags();
+    // Updating the tags by appending the keys of tags with a "*"
+    List<Tag> updateTagsList =
+        createdTagsList.stream()
+            .map(tag -> Tag.newBuilder().setId(tag.getId()).setKey(tag.getKey() + "*").build())
```
saxenakshitiz, line 102: Please don't use wild-card. Use character or number

tags-config-service-impl/src/test/java/org/hypertrace/tag/config/service/TagConfigServiceImplTest.java
```diff
@@ -0,0 +1,132 @@
...
+    createdTagsList.stream()
+        .forEach(
+            tag -> {
+              tagConfigStub.deleteTag(DeleteTagRequest.newBuilder().setId(tag.getId()).build());
+              List<Tag> allTags =
+                  tagConfigStub.getTags(GetTagsRequest.newBuilder().build()).getTagsList();
+              assertEquals(false, allTags.contains(tag));
+            });
+    List<Tag> allTags = tagConfigStub.getTags(GetTagsRequest.newBuilder().build()).getTagsList();
+    assertEquals(Collections.emptyList(), allTags);
```
saxenakshitiz, line 130: assert on size to be 0

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
kamaleshnneerasa: Okay, so what should be the behaviour if there is no tag with that id and key?

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
kamaleshnneerasa: Yes, I will make this change.

tags-config-service-impl/src/test/java/org/hypertrace/tag/config/service/TagConfigServiceImplTest.java
kamaleshnneerasa: Sure.

tags-config-service-impl/src/test/java/org/hypertrace/tag/config/service/TagConfigServiceImplTest.java
kamaleshnneerasa: Okay.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
jcchavezs: I would say throw exception. Does this need to be public? The interface does not seem to include this.

saxenakshitiz: We need to return 404 status here. Let me check on that front.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
saxenakshitiz: Can you check behavior on config service end? Probably it is already handled there itself.

kamaleshnneerasa: If we try to getConfig() on an ID that is not present, it is throwing an exception. So I think all the methods will also throw an exception. We have two options
1. We can add contains(id)/has(id) functionality to config-service.
2. In Tag-Config-Service, we can do a retrieve all the tags and filter them based on ID.
Let me know which approach to take.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
kamaleshnneerasa: What about the behaviour of getTag(id) and if id is not present?

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
saxenakshitiz: If getConfig() throws exception, then we just need to ensure it goes all the way back to the caller.

I am assuming deleteConfig() will also throw an exception, and we should ensure it goes all the way to the caller.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
saxenakshitiz: Ignore comment about 404 status. I think config service will throw exception with Status.NOT_FOUND, and that is good enough.

kamaleshnneerasa: Config-Service throws an exception with Status.UNKNOWN.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/ConfigServiceCoordinatorImpl.java
```diff
@@ -0,0 +1,122 @@
...
+          UpsertConfigRequest.newBuilder()
+              .setResourceName(TAG_CONFIG_RESOURCE_NAME)
+              .setResourceNamespace(TAG_CONFIG_RESOURCE_NAMESPACE)
+              .setConfig(convertToGenericFromTag(tag))
+              .setContext(tag.getId())
+              .build();
+      UpsertConfigResponse upsertConfigResponse = upsertConfig(requestContext, upsertConfigRequest);
+      return convertToTagFromGeneric(upsertConfigResponse.getConfig());
+    } catch (Exception e) {
+      throw e;
```
saxenakshitiz, line 49: If you are just catching and throwing e, then why have try/catch block? Please remove it. That is the default behaviour.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    responseObserver.onNext(GetTagsResponse.newBuilder().addAllTags(tagList).build());
+    responseObserver.onCompleted();
+  }
+
+  @Override
+  public void updateTag(
+      UpdateTagRequest request, StreamObserver<UpdateTagResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      Tag updateTag = request.getTag();
```
saxenakshitiz, line 73: nit: updatedTag

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/ConfigServiceCoordinatorImpl.java
```diff
@@ -0,0 +1,122 @@
...
+    try {
+      GetConfigRequest getConfigRequest =
+          GetConfigRequest.newBuilder()
+              .setResourceName(TAG_CONFIG_RESOURCE_NAME)
+              .setResourceNamespace(TAG_CONFIG_RESOURCE_NAMESPACE)
+              .addContexts(tagId)
+              .build();
+      GetConfigResponse getConfigResponse = getConfig(requestContext, getConfigRequest);
+      return convertToTagFromGeneric(getConfigResponse.getConfig());
+    } catch (Exception e) {
```
saxenakshitiz, line 68: same as above

saxenakshitiz (approved):

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  @Override
+  public void updateTag(
+      UpdateTagRequest request, StreamObserver<UpdateTagResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      Tag updatedTagInReq = request.getTag();
+      configServiceCoordinator.getTag(requestContext, updatedTagInReq.getId());
+      Tag updatedTagInRes = configServiceCoordinator.upsertTag(requestContext, request.getTag());
```
saxenakshitiz, line 75: use updatedTagInReq instead of request.getTag()

saxenakshitiz (approved):

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+import org.hypertrace.tag.config.service.v1.GetTagResponse;
+import org.hypertrace.tag.config.service.v1.GetTagsRequest;
+import org.hypertrace.tag.config.service.v1.GetTagsResponse;
+import org.hypertrace.tag.config.service.v1.Tag;
+import org.hypertrace.tag.config.service.v1.TagConfigServiceGrpc;
+import org.hypertrace.tag.config.service.v1.UpdateTagRequest;
+import org.hypertrace.tag.config.service.v1.UpdateTagResponse;
+
+@Slf4j
+public class TagConfigServiceImpl extends TagConfigServiceGrpc.TagConfigServiceImplBase {
```
jcchavezs, line 24: nit: this could be final.

tags-config-service-impl/src/main/java/org/hypertrace/tag/config/service/TagConfigServiceImpl.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  @Override
+  public void deleteTag(
+      DeleteTagRequest request, StreamObserver<DeleteTagResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      String tagId = request.getId();
+      configServiceCoordinator.getTag(requestContext, tagId);
```
jcchavezs, line 89: I am curious about this one. Do we truly need to check if tag exists before delete it? This operations don't seem to happen transactionally so after you get a tag and before you delete it, tag might be gone already. If delete is idempotent (which I believe it should), then I would just attempt to delete.

saxenakshitiz: @kamaleshnneerasa mentioned that there was a NPE when you attempt to delete a non-existent tag-id. You are right that we should fix it appropriately in root config service such that all config services are benefitted by this change.

PR #53, opened 2021-09-01
Added support for empty context in delete API

singhalprerana:
Added DEFAULT_CONTEXT as default when context is empty to delete corresponding config

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -5,5 +5,5 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
-        expires: 2021-09-02T00:00:00.000Z
+        expires: 2021-12-31T00:00:00.000Z
 patch: {}
--- a/.snyk
+++ b/.snyk
@@ -5,5 +5,5 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
-        expires: 2021-08-31T00:00:00.000Z
+        expires: 2021-09-02T00:00:00.000Z
 patch: {}
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -99,7 +99,8 @@ message DeleteConfigRequest {
   // required - namespace with which the config resource is associated
   string resource_namespace = 2;
 
-  // required
+  // optional - only required if config applies to a specific context.
+  // If empty, specified config is associated with a default context.
   string context = 3;
 }
 
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
@@ -156,7 +156,7 @@ public class ConfigServiceGrpcImpl extends ConfigServiceGrpc.ConfigServiceImplBa
         deleteConfigRequest.getResourceName(),
         deleteConfigRequest.getResourceNamespace(),
         getTenantId(),
-        deleteConfigRequest.getContext());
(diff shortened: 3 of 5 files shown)


aaron-steinfeld (commented):
Since it's failing on the build and only a one line fix, do you mind extending the expiration on the snyk ignore in `.snyk` ?

LGTM - @GurtejSohi just to confirm, there was no reason this was missing other than we didn't expect to want to do this, right?

singhalprerana:
> Since it's failing on the build and only a one line fix, do you mind extending the expiration on the snyk ignore in `.snyk` ?
Done.. but what is this exactly? if we are anyways ignoring, why add it in first place?

aaron-steinfeld:
> > Since it's failing on the build and only a one line fix, do you mind extending the expiration on the snyk ignore in `.snyk` ?
> 
> Done.. but what is this exactly? if we are anyways ignoring, why add it in first place?

So our only two options are to ignore it or fix it. So we don't permanently ignore it because it is a vulnerability we'd like to fix - https://snyk.io/vuln/SNYK-JAVA-IONETTY-1042268 , but can't do so yet as there's no fix available for the upstream (sounds from their github that it won't be until the next major release). The expiration is kind of a reminder for us to periodically check to see if there is a fix available - we usually push it out 3ish months at a time.

.snyk
```diff
@@ -5,5 +5,5 @@ ignore:
   SNYK-JAVA-IONETTY-1042268:
     - '*':
         reason: No replacement available
-        expires: 2021-08-31T00:00:00.000Z
+        expires: 2021-09-02T00:00:00.000Z
```
aaron-steinfeld, line 8:
```suggestion
        expires: 2021-12-31T00:00:00.000Z
```

jcchavezs:
@singhalprerana Just curious about why this is needed, in which cases we have such context?

singhalprerana:
> @singhalprerana Just curious about why this is needed, in which cases we have such context?

We store configs w.r.t. customers - we also want to delete them given customer use-cases like reset to default (e.g. new risk scoring we are currently working on)... Current API allows the config to gets stored wrt default_context. But there is no way to delete it..

GurtejSohi:
> LGTM - @GurtejSohi just to confirm, there was no reason this was missing other than we didn't expect to want to do this, right?

That's right. With this change, we should also update [this documentation comment](https://github.com/hypertrace/config-service/blob/main/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto#L102) - now the context is optional. @singhalprerana can you please update this?

aaron-steinfeld:
> > @singhalprerana Just curious about why this is needed, in which cases we have such context?
> 
> We store configs w.r.t. customers - we also want to delete them given customer use-cases like reset to default (e.g. new risk scoring we are currently working on)... Current API allows the config to gets stored wrt default_context. But there is no way to delete it..

Slightly more broadly - some config has its own ID (imagine something like user preferences), other config could be global to the tenant (such as something like agent configuration). In the config model, that global scenario means it has a resource name/namespace, but no specific context - it's global. As @singhalprerana mentioned, we can write to and read that tenant-wide config, but we can't remove it. I can't recall if that's intentional, assuming that we'd never want to have no config, or just a bug, but even if it was intentional, it doesn't match how we often use it for this type of scenario. That is, loading a default value from config, and only storing an override - so a "reset to default" use case would involve deleting the override.

singhalprerana:
> > LGTM - @GurtejSohi just to confirm, there was no reason this was missing other than we didn't expect to want to do this, right?
> 
> That's right. With this change, we should also update [this documentation comment](https://github.com/hypertrace/config-service/blob/main/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto#L102) - now the context is optional. @singhalprerana can you please update this?

done

aaron-steinfeld (approved):

GurtejSohi (approved):

PR #56, opened 2021-09-14
Adding APIs for label application rule

saxenakshitiz:


label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto
```diff
@@ -0,0 +1,64 @@
...
+  // update label application rule
+  rpc UpdateLabelApplicationRule(UpdateLabelApplicationRuleRequest) returns (UpdateLabelApplicationRuleResponse) {}
+
+  // delete label application rule by id
+  rpc DeleteLabelApplicationRule(DeleteLabelApplicationRuleRequest) returns (DeleteLabelApplicationRuleResponse) {}
+
+}
+
+message CreateLabelApplicationRuleRequest {
+  CreateLabelApplicationRule label_application_rule = 1;
```
aaron-steinfeld, line 29: We already have a create-specific `CreateLabelApplicationRuleRequest`, we don't need `CreateLabelApplicationRule`, we can just unwrap it and have a `LabelApplicationRuleData`

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto
```diff
@@ -0,0 +1,64 @@
...
+
+message GetLabelApplicationRulesRequest {
+}
+
+message GetLabelApplicationRulesResponse {
+  repeated LabelApplicationRule llabel_application_rules = 1;
+}
+
+message UpdateLabelApplicationRuleRequest {
+  LabelApplicationRule label_application_rule = 1;
```
aaron-steinfeld, line 52: By separating this out here, we leave room to add read-only fields in the future. For example, imagine adding a creation date field to the rule, which we would not want to support for editing.

e.g.
```proto
message UpdateLabelApplicationRuleRequest {
  string id = 1;
  LabelApplicationRule data = 2;
}
```

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,98 @@
...
+  LabelApplicationRule data = 2;
+}
+
+message CreateLabelApplicationRule {
+  LabelApplicationRule data = 2;
+}
+
+message LabelApplicationRuleData {
+  string name = 1;
+  string description = 2;
```
aaron-steinfeld, line 18: Not sure what our current requirements are, but I find we never really use description  and name is often just a projection of the data inside (e.g. "Apply 'Sensitive' Label"). If we don't have a requirement for user-specified name and description, I'd drop em. At least description, since if we're allowing a custom name that serves as the description.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,98 @@
...
+    repeated Condition conditions = 2;
+
+    enum Type {
+      TYPE_UNSPECIFIED = 0;
+      TYPE_SPAN = 1;
+      TYPE_TRACE = 2;
+    }
+
+    message Condition {
+      string name = 1;
```
aaron-steinfeld, line 33: What's name?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,98 @@
...
+      Scope scope = 2;
+      string key = 3;
+      Operator operator = 4;
+      optional string value = 5;
+
+      enum Scope {
+        SCOPE_UNSPECIFIED = 0;
+        SCOPE_ATTRIBUTE = 1;
+        SCOPE_REQUEST = 2;
```
aaron-steinfeld, line 42: request and response are just attributes (well, requestBody/requestHeaders/responseBody/responseHeaders are). Do we need to separate them? This also differs from how we use "scope" elsewhere in the platform and in this PR.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,98 @@
...
+        SCOPE_RESPONSE = 3;
+      }
+
+      enum Operator {
+        OPERATOR_UNSPECIFIED = 0;
+        OPERATOR_EXISTS = 1;
+        OPERATOR_EQUALS = 2;
+        OPERATOR_CONTAINS = 3;
+        OPERATOR_STARTS_WITH = 4;
```
aaron-steinfeld, line 51: What's contains vs exists? Is contains operating on the value but exists only on the key with value being ignored? Operator names probably could use some clarification. Also, aren't contains/starts with/ends with just regex cases - do we need to support them as top level operators?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,98 @@
...
+      TYPE_SPAN = 1;
+      TYPE_TRACE = 2;
+    }
+
+    message Condition {
+      string name = 1;
+      Scope scope = 2;
+      string key = 3;
+      Operator operator = 4;
+      optional string value = 5;
```
aaron-steinfeld, line 37: Can values be of different data types?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,98 @@
...
+    Type type = 1;
+    Scope scope = 2;
+    Action action = 3;
+
+    enum Type {
+      TYPE_UNSPECIFIED = 0;
+      TYPE_ENTITY = 1;
+    }
+
+    enum Scope {
```
aaron-steinfeld, line 68: Can we abstract this so it's not an enum? That might just mean replacing it with a string. The actual entities support in the system are config driven, and should remain so here.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,98 @@
...
+        OPERATION_UNSPECIFIED = 0;
+        OPERATION_MERGE = 1;
+      }
+
+      message DynamicLabel {
+        repeated TokenExtractionRule token_extraction_rules = 1;
+        string token_combining_expression  = 2;
+
+        message TokenExtractionRule {
+          string key = 1;
```
aaron-steinfeld, line 92: Are these span keys? trace keys?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,98 @@
...
+      }
+
+      enum Operation {
+        OPERATION_UNSPECIFIED = 0;
+        OPERATION_MERGE = 1;
+      }
+
+      message DynamicLabel {
+        repeated TokenExtractionRule token_extraction_rules = 1;
+        string token_combining_expression  = 2;
```
aaron-steinfeld, line 89: What's the expression syntax? Do we even need the token rules if alias is optional?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto
```diff
@@ -0,0 +1,64 @@
...
+
+message GetLabelApplicationRuleResponse {
+  LabelApplicationRule label_application_rule = 1;
+}
+
+message GetLabelApplicationRulesRequest {
+}
+
+message GetLabelApplicationRulesResponse {
+  repeated LabelApplicationRule llabel_application_rules = 1;
```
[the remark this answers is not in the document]
saxenakshitiz: Done

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto
saxenakshitiz: Sure

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto
saxenakshitiz: I didn't fully get this. In config service, we auto populate create and update timestamp. This is not even exposed as part of individual service config.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Yes, description is generally not used. I will remove it. I will be keep the name so that user can give the rule a relevant name.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Condition name like `Condition 1` 
Depends how it will be displayed in UI.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: I thought it is simplest to start with string data type. We can add support for more data types for values as and when needed. It will make out operator support richer. For numbers, we can support `<`, `>`  etc. 

What do you say? Or shall we add support for numbers right now?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Even though your point is valid, not user whether end user will prefer that distinction. A single bucket of attributes is simpler in terms of implementation. Will check with Sudeep on this front.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: You have captured the difference between exists and contains correctly. Let me add comments to provide clarity.

For string values will like to support contains, ends-with and starts-with as first class operators. This is similar to what we have java.lang.String.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: okay. However we will have validation around it to ensure that the entity that is being labelled is either API, Service or Domain. Correct?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: I was thinking of format supported by slf4j MessageFormatter

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Will be picked from type defined in matching criteria?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: We can get rid of token extraction rule completely and just have `label_expression` which can be of format `${deployment}_${version}`. I forgot to add whether we need to extract token from request/response/attribute. If we are just going with attribute, then `TokenExtractionRule` is probably not needed.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto
aaron-steinfeld: We don't expose these today, create timestamp was just being used as an example of a field we might want defined on a rule, but never exposed for either create or update.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: I would expect we'd display the condition fields themselves - I don't think there's any need to name a sub-object of a rule.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: Yes there's the end user question, but that's not necessarily something we need to capture at this level. First and foremost, we want this to be flexible, so we need not build out special cases for ease of use - we'd rather be generic and handle the ease of use at a higher layer.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: Is there a requirement somewhere for those operators? If not, I'd punt and can always add later. The difference in java is that convenience is a one-level thing - it's syntax sugar. Here, the addition of each operation spans many different parts of the system, so they're fairly expensive and we should be conservative about the ones we expose. The application of the label needs to know about it (implementation), the external API needs to support exposing it (serialization), the UI needs to support building it (presentation) etc.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: Ideally, no - at least not in code. A runtime check on the other hand could be feasible by a call to entity type system (but config service shouldn't be calling there) to ensure an entity type exists. Not every system will have every entity so we don't want to bake in any assumption if we can help it - either in api or impl.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: I think this is where part of the confusion of scope comes in, too. What's being called type in matching criteria is scope elsewhere in the system. "API", "SERVICE", "SPAN", "TRACE" - all scopes.

Further, we're designed to enforce a single scope across all conditions. Do we think it will ever be a requirement to - for example - want a trace to match condition X and a span to match condition Y? I could definitely envision that, as the trace gives data from an end user perspective (e.g. what was the host header on the external call), while the span gives more specific data (what service is this, what data was passed in etc.).

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: So the above example would apply the label "mydeployment-1.3" if the span had below attributes? Just want to make sure I understand correctly:
```json
{
  "deployment": "mydeployment",
  "version": "1.3"
}
```

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: We don't need to add number support right now, but it gets at the design evolution. The operators for strings and numbers are different (as are for unary, no-arg operators like `exists`, for that matter). So if we were to add number support, would we be doing something like a oneof with different messages encompassing the string operators/value, numeric operators/value, unary etc?) I'm less concerned with functionality than compatibility.

```proto
message Condition {
  oneof condition {
    StringCondition string_condition = 1;
    NumericCondition numeric_condition = 2;
    UnaryCondition unary_condition = 3;
  }
}
  ```

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 4 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto:28]: Removed

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto (hunk 2 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto:52]: Sure. I don't have a problem with this approach. Updated accordingly.

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 5 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: If we don't capture user's input as is in config, then we will need to transform both ways - from user input to what's persisted in config, as well as config to what's presented to user.

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 6 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: I don't think we have any specific requirement around operator. What do you think we need to support at minimum?

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 10 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: Yes

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 7 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto:32]: Makes sense. Updated accordingly

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 8 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: Runtime check is a bit too late. We need to notify user if he/she has selected a wrong scope?

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 9 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto:73]: I have defined type and scope. Type can be `span or trace` for matching criteria, and `entity` for application criteria. Source can be `request, response or attribute` for key. If this terminology is confusing, then we can change it.

Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto:73]: There may be cases where heterogeneous conditions support is required.  To support  heterogeneous conditions, we can have `type` at `condition level`. Similarly we will need to expose `type` in `token extraction rule` as well.

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 8 above)
Comment (aaron-steinfeld) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: By runtime, I mean during validation rather than execution. The distinction I was making with runtime was looking up the allowed scope strings vs defining them in code.

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 6 above)
Comment (aaron-steinfeld) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: I'd probably just start with exists/equals/regex

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 5 above)
Comment (aaron-steinfeld) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: We'll have to transform either way, since we're not presenting protos to users :)

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 9 above)
Comment (aaron-steinfeld) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto:84]: I think potentially prefixing the words like `target_scope`, `matching_scope` etc. would make sense if all of the values are subsets of our general scope (which span, trace and entities are). WRT request/response/attribute, I'd use a different word there if we preserve it, but I'd hope to get away from it as discussed on a different thread.

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 6 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: Done

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto (hunk 8 above)
Comment (saxenakshitiz) [reply label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto]: Decision here is that validation will be based on supported entity types provided in config in short term.

In the long term, entity type service should provide details about which entity type can be labelled. Once this information is available, it can be fetched at any place for validation purpose. However config service will not pull this detail for validation from entity type service, as it may bring cyclic dependency. The validation will happen at higher layers like UI, GraphQL.

File: label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto [hunk 12]
```diff
@@ -0,0 +1,104 @@
...
+
+message LabelApplicationRule {
+  string id = 1;
+  LabelApplicationRuleData data = 2;
+}
+
+message LabelApplicationRuleData {
+  string name = 1;
+  MatchingCriteria matching_criteria = 2;
+  ApplicationCriteria application_criteria = 3;
```
aaron-steinfeld, line 15: is this one really a criteria? Think this naming could be tweaked

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,104 @@
...
+
+  message MatchingCriteria {
+    Condition condition = 1;
```
aaron-steinfeld, line 18: If we're down to a single field in this, we can probably drop a layer of envelopes.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,104 @@
...
+
+      enum LogicalOperator {
+        LOGICAL_OPERATOR_UNSPECIFIED = 0;
+        LOGICAL_OPERATOR_AND = 1;
+        LOGICAL_OPERATOR_OR = 2;
+      }
+    }
+
+    message LeafCondition {
+      Scope scope = 1;
```
aaron-steinfeld, line 39: mentioned offline, no longer need scope

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,104 @@
...
+
+      enum Operator {
+        OPERATOR_UNSPECIFIED = 0;
+        OPERATOR_EQUALS = 1; // operator to check if the key exists, and value is equal to the provided value
+        OPERATOR_MATCHES_REGEX = 2; // operator to check if the key exists, and value matches the provided regex value
+      }
+    }
+
+    message UnaryCondition {
+      // used to check the existence of the key
```
aaron-steinfeld, line 60: may want to still have an operator here as there may be multiple unary conditions in the future (e.g. exists and not exists)

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,96 @@
...
+
+      enum LogicalOperator {
+        LOGICAL_OPERATOR_UNSPECIFIED = 0;
+        LOGICAL_OPERATOR_AND = 1;
+        LOGICAL_OPERATOR_OR = 2;
+      }
+    }
+
+    message LeafCondition {
+      string key = 1;
```
aaron-steinfeld, line 39: We may want to specify this as a condition too based on the conversations. Adding regex support on the key here would require a breaking change, but something like below would support it.


```proto
    message LeafCondition {
      StringCondition key_condition = 1;
      oneof value_condition {
        StringCondition string_value_condition = 2;
        UnaryCondition unary_value_condition = 3;
        JsonCondition json_value_condition = 4;
      }
    }
```

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Done

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Condition is used recursively in CompositeCondition. Let's keep matching and application criteria?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: This makes sense. However do note that key in condition can be regex based, however for dynamic label we will require specific key. Isn't it?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: What about changing line 14 to `Condition matching_criteria = 2;` That shouldn't get in the way of the recursion. I'm fine either way though. More than anything, I just want application criteria to be renamed since it _isn't_ a criteria.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
aaron-steinfeld: Given that the label has its own specification of source I don't think any labeling concerns are relevant to the conditions? Unless I'm misunderstanding?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,104 @@
...
+      }
+    }
+
+    message StringCondition {
+      Operator operator = 1;
+      string value = 2;
+
+      enum Operator {
+        OPERATOR_UNSPECIFIED = 0;
+        OPERATOR_EQUALS = 1; // operator to check if the key exists, and value is equal to the provided value
```
surajpuvvada, line 53: is the comment correct ? there is no key here. ?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
surajpuvvada: yea Condition matching_criteria is more readable but ApplicationCriteria is confusing.A rule essentially has conditions and actions so may be Actions ? ( I see Action is already used)

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -0,0 +1,104 @@
...
+
+  message ApplicationCriteria {
+    string entity_type = 1; // captures the entity type - "API", "Service"
+    Action action = 2;
+
+    message Action {
+      Operation operation = 1;
+      oneof value {
+        string static_label_id = 2;
+        DynamicLabel dynamic_label = 3;
```
surajpuvvada, line 84: so dynamic labels are not defined in the system with an id ? what is the difference between static_label_id and dynamic_label ?

surajpuvvada:
Are we going to have label extraction rules as well ?

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Removed application_criteria and directly used action.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Removed both criteria, and used condition and action directly.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Key is captured at leaf condition level

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: All the labels will be associated with an entity using an id. That is same for static or dynamic labels.

Please note that static labels pre-exists in the system and will have id already associated with it. Dynamic labels are created at runtime. If it doesn't exist, then it is created along with its id. Being dynamic in nature, the api definition only captures how to generate them.

aaron-steinfeld:
> Are we going to have label extraction rules as well ?

what do you mean by label extraction? Is that something different than the DynamicLabel message?

Overall, lgtm.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Removed

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
saxenakshitiz: Not a concern. Just an observation.

surajpuvvada (dismissed):
lgtm

aaron-steinfeld (dismissed):
please fix snyk dependency issue here or separately.

aaron-steinfeld (approved):

PR #55, opened 2021-09-07
feat: HT Alerting config service

rish691:
https://github.com/hypertrace/hypertrace/issues/300 

Config service to persist a generic alert rule in document store

PR #57, opened 2021-09-17
feat: Notification config service

rish691:
https://github.com/hypertrace/hypertrace/issues/301

Config service for persisting notification rule and channel in document store


notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
```diff
@@ -0,0 +1,30 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+
+package org.hypertrace.notification.config.service.v1;
+
+message NewNotificationRule {
+  string rule_name = 1;
+  string description = 2;
+  string environment = 3;
```
aaron-steinfeld, line 10: environment isn't really relevant here. Should this support be moved into an event condition rather than first classing it?

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceStore.java
```diff
@@ -0,0 +1,205 @@
+package org.hypertrace.notification.config.service;
```
aaron-steinfeld, line 1: This whole class should be replaceable by the object store PR

rish691:
Will update this pr once https://github.com/hypertrace/config-service/pull/55 goes through

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
rish691: > environment isn't really relevant here. 
why?

tbh I just copied it from the previous proto, do you mean it should be in EventCondition proto in the other pr?

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -0,0 +1,39 @@
...
+  string id = 1;
+  NotificationChannelMutableData notification_channel_mutable_data = 2;
+}
+
+message NotificationChannelMutableData {
+  string channel_name = 1;
+  NotificationChannelConfig notification_channel_config = 2;
+}
+
+message NotificationChannelConfig {
```
aaron-steinfeld, line 21: can this collapse into the mutabledata that contains it?

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
```diff
@@ -0,0 +1,25 @@
...
+message NotificationRule {
+  string id = 1;
+  NotificationRuleMutableData notification_rule_data = 2;
+}
+
+message NotificationRuleMutableData {
+  string rule_name = 1;
+  string description = 2;
+
+  string event_condition_id = 3;
```
aaron-steinfeld, line 20: is every notification rule associated with exactly one event condition? Or could it it have multiple? e.g. X and Y

notification-config-service-impl/build.gradle.kts
```diff
@@ -0,0 +1,34 @@
...
+
+dependencies {
+    api(projects.notificationConfigServiceApi)
+    implementation(projects.configServiceApi)
+    implementation(projects.configProtoConverter)
+    implementation(projects.objectStore)
+    implementation(projects.validationUtils)
+
+    implementation(libs.guava)
+    implementation(libs.rxjava3)
```
aaron-steinfeld, line 15: make sure we're only declaring libraries being used, don't see rx being used.

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelStore.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  protected Optional<NotificationChannel> buildObjectFromValue(Value value) {
+    NotificationChannel.Builder builder = NotificationChannel.newBuilder();
+    try {
+      ConfigProtoConverter.mergeFromValue(value, builder);
+      return Optional.of(builder.build());
+    } catch (InvalidProtocolBufferException e) {
+      log.error("Conversion failed. value {}", value, e);
+      throw new RuntimeException(e);
```
aaron-steinfeld, line 38: Should return optional empty here. The IdentifiedObjectStore handles this, either throwing if it's a get by id/upsert, or dropping if it's getAll so a single bad value doesn't break the resource entirely.

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceConstants.java
```diff
@@ -0,0 +1,7 @@
+package org.hypertrace.notification.config.service;
+
+interface NotificationConfigServiceConstants {
+  String NOTIFICATION_RULE_CONFIG_RESOURCE_NAME = "notificationRuleConfig";
```
aaron-steinfeld, line 4: Maybe the namespace should be shared since we have two different resources here, but the actual resource names are internal details to the specific store and need not have exposed constants.

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceImpl.java
```diff
@@ -0,0 +1,230 @@
...
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      notificationConfigServiceRequestValidator.validateCreateNotificationRuleRequest(
+          requestContext, request);
+      NewNotificationRule newNotificationRule = request.getNewNotificationRule();
+      NotificationRule.Builder builder =
+          NotificationRule.newBuilder()
+              .setId(UUID.randomUUID().toString())
+              .setNotificationRuleData(
+                  NotificationRuleMutableData.newBuilder()
```
aaron-steinfeld, line 59: This no longer needs the mapping with the mutable data approach right? Can we set `setNotificationRuleData(newNotificationRule.getNotificationRuleData())`

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -0,0 +1,39 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+
+package org.hypertrace.notification.config.service.v1;
+
+message NewNotificationChannel {
```
aaron-steinfeld, line 7: These extra wrappers can go away with the mutable data pattern. Instead, update the create/update requests to take in mutable data and mutable data + id, respectively.

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceImpl.java
```diff
@@ -0,0 +1,230 @@
...
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      notificationConfigServiceRequestValidator.validateCreateNotificationChannelRequest(
+          requestContext, request);
+      NotificationChannelMutableData notificationChannelMutableData =
+          request.getNewNotificationChannel().getNotificationChannelMutableData();
+      NotificationChannel.Builder builder =
+          NotificationChannel.newBuilder()
+              .setId(UUID.randomUUID().toString())
+              .setNotificationChannelMutableData(
+                  NotificationChannelMutableData.newBuilder()
```
aaron-steinfeld, line 155: Here too

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceImpl.java
```diff
@@ -0,0 +1,230 @@
...
+      StreamObserver<CreateNotificationChannelResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      notificationConfigServiceRequestValidator.validateCreateNotificationChannelRequest(
+          requestContext, request);
+      NotificationChannelMutableData notificationChannelMutableData =
+          request.getNewNotificationChannel().getNotificationChannelMutableData();
+      NotificationChannel.Builder builder =
+          NotificationChannel.newBuilder()
+              .setId(UUID.randomUUID().toString())
```
aaron-steinfeld, line 153: Nit - we have a uuidgenerator util too (useful for testing to inject rather than use a static)

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,111 @@
...
+import org.hypertrace.notification.config.service.v1.GetAllNotificationRulesRequest;
+import org.hypertrace.notification.config.service.v1.UpdateNotificationChannelRequest;
+import org.hypertrace.notification.config.service.v1.UpdateNotificationRuleRequest;
+
+public class NotificationConfigServiceRequestValidator {
+
+  public void validateCreateNotificationRuleRequest(
+      RequestContext requestContext, CreateNotificationRuleRequest request) {
+    validateRequestContextOrThrow(requestContext);
+    Preconditions.checkArgument(
```
aaron-steinfeld, line 21: These should switch to use common validator utils (less code and better error messaging).

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,111 @@
...
+        !request.getNewNotificationRule().getNotificationRuleData().getChannelId().isBlank(),
+        "ChannelId cannot be empty");
+  }
+
+  public void validateUpdateNotificationRuleRequest(
+      RequestContext requestContext, UpdateNotificationRuleRequest request) {
+    validateRequestContextOrThrow(requestContext);
+    Preconditions.checkArgument(
+        request.hasNotificationRule(), "Notification rule should be present");
+    Preconditions.checkArgument(
```
aaron-steinfeld, line 36: update and create can share the mutable data validation

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleStore.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  protected Optional<NotificationRule> buildObjectFromValue(Value value) {
+    NotificationRule.Builder builder = NotificationRule.newBuilder();
+    try {
+      ConfigProtoConverter.mergeFromValue(value, builder);
+      return Optional.of(builder.build());
+    } catch (InvalidProtocolBufferException e) {
+      log.error("Conversion failed. value {}", value, e);
+      throw new RuntimeException(e);
```
aaron-steinfeld, line 38: same comment here

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -0,0 +1,39 @@
...
+  NotificationChannelMutableData notification_channel_mutable_data = 1;
+}
+
+message NotificationChannel {
+  string id = 1;
+  NotificationChannelMutableData notification_channel_mutable_data = 2;
+}
+
+message NotificationChannelMutableData {
```
kotharironak, line 16: Note: I think, we will need changes in `notification-channels.json` accordingly in hypertrace-alert-engine, right? If, yes, can we log an issue for that as well?

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_config_service.proto
```diff
@@ -0,0 +1,85 @@
...
+  NotificationChannel notification_channel = 1;
+}
+
+message GetAllNotificationChannelsRequest {}
+
+message GetAllNotificationChannelsResponse {
+  repeated NotificationChannel notification_channels = 1;
+}
+
+message GetNotificationChannelRequest {
```
kotharironak, line 73: I don't see a corresponding `Get` request based on id as part of APIs for channel. So, is this requires or do we have to add get request?

notification-config-service-impl/build.gradle.kts
kotharironak: +1

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_config_service.proto
```diff
@@ -0,0 +1,85 @@
...
+import "org/hypertrace/notification/config/service/v1/notification_rule.proto";
+import "org/hypertrace/notification/config/service/v1/notification_channel.proto";
+
+service NotificationConfigService {
+  rpc CreateNotificationRule(CreateNotificationRuleRequest) returns (CreateNotificationRuleResponse) {}
+  rpc UpdateNotificationRule(UpdateNotificationRuleRequest) returns (UpdateNotificationRuleResponse) {}
+  rpc GetAllNotificationRules(GetAllNotificationRulesRequest) returns (GetAllNotificationRulesResponse) {}
+  rpc DeleteNotificationRule(DeleteNotificationRuleRequest) returns (DeleteNotificationRuleResponse) {}
+
+  rpc CreateNotificationChannel(CreateNotificationChannelRequest) returns (CreateNotificationChannelResponse) {}
```
kotharironak, line 16: Should the Notification channel be a separate service - `NotificationChannelServiceImpl`? The reason being the extension of HT might have support for more channels. In that case, a different overridden service can be registered in the config service factory. Till there is a common set, it can continue to use HT one.

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelStore.java
kotharironak, line 38: Should return be `Optional.empty();`? - https://github.com/hypertrace/config-service/blob/main/alerting-config-service-impl/src/main/java/org/hypertrace/alerting/config/service/EventConditionStore.java#L39?

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,111 @@
...
+
+  public void validateCreateNotificationRuleRequest(
+      RequestContext requestContext, CreateNotificationRuleRequest request) {
+    validateRequestContextOrThrow(requestContext);
+    Preconditions.checkArgument(
+        request.hasNewNotificationRule(), "Notification rule should be present");
+    Preconditions.checkArgument(
+        !request.getNewNotificationRule().getNotificationRuleData().getRuleName().isBlank(),
+        "Rule name cannot be empty");
+    Preconditions.checkArgument(
```
kotharironak, line 26: Can event_condition_id and type be empty? Do we need to check for that too?

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
kotharironak: Good point. 

These are independent events. So, for a rule - send a notification if both event X and event Y occurred - notifier will need more information for evaluation (How long should wait for both events to occur? During what time-window should it consider for evaluation?). 

Instead of going via notifier, as the second way, we can consider such rules as a new event condition type with its evaluator.  So, I think we can go with one-one mapping.

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_config_service.proto
rish691: Thanks for pointing out, this isn't needed at the moment, removed.

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceImpl.java
rish691: that is not present in ht ig

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
rish691: I am keeping mutable data + id as a message, the create request works with mutable data

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceRequestValidator.java
rish691: But the utils can't be used to check the presence of message type of field?

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
aaron-steinfeld: Yeah makes sense. I was thinking of cases that multiple conditions apply to the same type of event, but those likely belong to a single top level condition that may contain multiple predicates (e.g. a metric that has name X, has violated a baseline Y and belongs to API Z).

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceImpl.java
aaron-steinfeld: ah so it's not. I'll promote it when I get a chance.

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
aaron-steinfeld: it makes sense to have mutable + id as a message on the return side (that is, we're returning a NotificationChannel whether we get, create or update), but on the request side, we're lumping two different concerns into the same message. The `NotificationChannel` represents the full state of a channel, which today is an ID + mutable data. What happens when a second immutable field is added, say last edit date or created date? That belongs to the rule being returned, but it doesn't belong to the update API - the update API should always be taking just an identifier and the data to change.

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceRequestValidator.java
aaron-steinfeld: This is by design - checking `hasNewNotificationRule` here is redundant with the later checks checking that values inside are valid. We validate on primitives (that is, leaf nodes of the message), so if we don't require any of the primitives inside a message, there's no reason to require the message.

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_config_service.proto
aaron-steinfeld: yeah, makes sense to go granular.

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
rish691: But the object returned from get and used in upsert is assumed to be same in ObjectStore abstraction?

notification-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationConfigServiceRequestValidator.java
rish691: yeah makes sense, if primitive is not required then the message is not required.

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_config_service.proto
rish691: Separate out notification-channel in a separate module (keeping in the same pr as most of the review is done)

notification-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
aaron-steinfeld: At the object store level, it works in the same objects across all APIs. From perspective of service:

Get - lookup NotificationChannel by id via store, return NotificationChannel
Update - take in id and mutable data, lookup corresponding NotificationChannel by id via store, merge changes and upsert via store. Return NotificationChannel.
Create - use mutable data to create a new NotificationChannel with generated ID and upsert via store.


So Update may be doing an extra lookup right now - but that's something we can shortcut in the implementation (just use that data to create a NotificationChannel if we have every single field and no merging is needed). By keeping that shortcut out of the API though, we can support either - a partial update where only mutable fields are changing, or a full update where we are accepting every single field. Does that make sense?

notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,51 @@
...
+        request.getNotificationChannel(), NotificationChannel.ID_FIELD_NUMBER);
+    validateNotificationChannelMutableData(
+        request.getNotificationChannel().getNotificationChannelMutableData());
+  }
+
+  private void validateNotificationChannelMutableData(NotificationChannelMutableData data) {
+    validateNonDefaultPresenceOrThrow(
+        data, NotificationChannelMutableData.CHANNEL_NAME_FIELD_NUMBER);
+    Preconditions.checkArgument(
+        data.getEmailChannelConfigCount() != 0 || data.getWebhookChannelConfigCount() != 0,
```
aaron-steinfeld, line 36: The logic makes sense, but can we throw this as an invalid arg status exception.

aaron-steinfeld: Although, should we be checking each email/webhook config for validity, too?

rish691: Ideally yes, but previously there wasn't any check. 

Basically we need to validate the url and email address. Probably use some util to do that. I'll maybe do it later on?

aaron-steinfeld: Sure

aaron-steinfeld (approved):

PR #60, opened 2021-09-22
CRUD API implementation for Label Application Rule Config Service

kamaleshnneerasa:
## Description

### Testing

### Checklist:

### Documentation


label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -0,0 +1,164 @@
...
+import org.hypertrace.label.application.rule.config.service.v1.GetLabelApplicationRulesRequest;
+import org.hypertrace.label.application.rule.config.service.v1.GetLabelApplicationRulesResponse;
+import org.hypertrace.label.application.rule.config.service.v1.LabelApplicationRule;
+import org.hypertrace.label.application.rule.config.service.v1.LabelApplicationRuleConfigServiceGrpc;
+import org.hypertrace.label.application.rule.config.service.v1.UpdateLabelApplicationRuleRequest;
+import org.hypertrace.label.application.rule.config.service.v1.UpdateLabelApplicationRuleResponse;
+
+public class LabelApplicationRuleConfigServiceImpl
+    extends LabelApplicationRuleConfigServiceGrpc.LabelApplicationRuleConfigServiceImplBase {
+  private static final int LABEL_LOCK_STRIPE_COUNT = 1000;
```
saxenakshitiz, line 29: They are not used currently. Plz remove them.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -0,0 +1,143 @@
...
+    if (compositeCondition.getOperator()
+        == CompositeCondition.LogicalOperator.LOGICAL_OPERATOR_UNSPECIFIED) {
+      throw new IllegalArgumentException("Invalid Logical Operator");
+    }
+    compositeCondition.getChildrenList().forEach(this::validateCondition);
+  }
+
+  private void validateJsonCondition(JsonCondition jsonCondition) {
+    if (!jsonCondition.hasStringCondition() && !jsonCondition.hasUnaryCondition()) {
+      throw new IllegalArgumentException("Condition in JSON Condition not set");
```
saxenakshitiz, line 113: Missing String or Unary condition in JSON Condition

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
kamaleshnneerasa: okay.

label-application-rule-config-service-impl/build.gradle.kts
```diff
@@ -0,0 +1,33 @@
...
+
+dependencies {
+    api(projects.labelApplicationRuleConfigServiceApi)
+    implementation(projects.configServiceApi)
+    implementation(projects.configProtoConverter)
+    implementation(libs.typesafe.config)
+    implementation(libs.protobuf.javautil)
+    implementation(libs.slf4j.api)
+    implementation(libs.guava)
+    implementation(libs.rxjava3)
```
aaron-steinfeld, line 15: please go through the dependencies and limit to what is needed - I checked a couple (rxjava and guava) and neither is used.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/ConfigServiceCoordinator.java
```diff
@@ -0,0 +1,16 @@
+package org.hypertrace.label.application.rule.config.service;
+
+import java.util.List;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.label.application.rule.config.service.v1.LabelApplicationRule;
+
+public interface ConfigServiceCoordinator {
```
aaron-steinfeld, line 7: This pattern has been abstracted in IdentifiedObjectStore, you should be able to use that close to directly.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceConstants.java
```diff
@@ -0,0 +1,9 @@
+package org.hypertrace.label.application.rule.config.service;
+
+public class LabelApplicationRuleConfigServiceConstants {
```
aaron-steinfeld, line 3: no separate constants class needed, these are private to the data abstraction.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  @Override
+  public void validateOrThrow(
+      RequestContext requestContext,
+      DeleteLabelApplicationRuleRequest deleteLabelApplicationRuleRequest) {
+    validateRequestContext(requestContext);
+  }
+
+  private void validateRequestContext(RequestContext requestContext) {
```
aaron-steinfeld, line 56: You can use the common validation utils to replace this and many of the specific errors below.

kamaleshnneerasa: Getting this error when calling the common validation apis on some fields "Caused by: java.lang.UnsupportedOperationException: FieldDescriptor.getDefaultValue() called on an embedded message field".

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/ConfigServiceCoordinator.java
kamaleshnneerasa: okay.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
aaron-steinfeld: What was the corresponding code? I believe you'll hit that if you try to validate a message (as opposed to a primitive) is set, but that's by design.

Take a message like 
```proto
message WrapperMessage {
  InnerMessage inner_message = 1;
 
  message InnerMessage {
     string str_value = 1;
  }
}
```

If you want to make sure that str_value is set, you'd check it directly - and that should work. Checking if InnerMessage is set doesn't really provide validation value and I believe that's what would throw today. We could add support for it given a compelling use case, but I've yet to see one.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+import org.hypertrace.label.application.rule.config.service.v1.GetLabelApplicationRulesRequest;
+import org.hypertrace.label.application.rule.config.service.v1.GetLabelApplicationRulesResponse;
+import org.hypertrace.label.application.rule.config.service.v1.LabelApplicationRule;
+import org.hypertrace.label.application.rule.config.service.v1.LabelApplicationRuleConfigServiceGrpc;
+import org.hypertrace.label.application.rule.config.service.v1.UpdateLabelApplicationRuleRequest;
+import org.hypertrace.label.application.rule.config.service.v1.UpdateLabelApplicationRuleResponse;
+
+public class LabelApplicationRuleConfigServiceImpl
+    extends LabelApplicationRuleConfigServiceGrpc.LabelApplicationRuleConfigServiceImplBase {
+  public static final String LABEL_APPLICATION_RULE_CONFIG_RESOURCE_NAME = "label-config";
```
aaron-steinfeld, line 30: private impl details, these can be inlined directly in `LabelApplicationRuleStore`

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+  @Override
+  public void createLabelApplicationRule(
+      CreateLabelApplicationRuleRequest request,
+      StreamObserver<CreateLabelApplicationRuleResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      this.requestValidator.validateOrThrow(requestContext, request);
+      LabelApplicationRule labelApplicationRule =
+          LabelApplicationRule.newBuilder()
+              .setId(UUID.randomUUID().toString())
```
aaron-steinfeld, line 58: nit/suggestion: Use `UuidGenerator` (which makes it easier to test, since that can be a mocked dependency)

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+              .setLabelApplicationRule(createdLabelApplicationRule)
+              .build());
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      responseObserver.onError(e);
+    }
+  }
+
+  @Override
+  public void getLabelApplicationRule(
```
aaron-steinfeld, line 74: Out of curiosity, do we have a get by ID use case today? If not, maybe we should just drop this API until/unless needed.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+      GetLabelApplicationRuleRequest request,
+      StreamObserver<GetLabelApplicationRuleResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      this.requestValidator.validateOrThrow(requestContext, request);
+      String ruleId = request.getId();
+      Optional<LabelApplicationRule> optionalLabelApplicationRule =
+          this.labelApplicationRuleStore.getObject(requestContext, ruleId);
+      if (optionalLabelApplicationRule.isEmpty()) {
+        responseObserver.onError(new StatusRuntimeException(Status.NOT_FOUND));
```
aaron-steinfeld, line 84: nit: `Status.NOT_FOUND.asRuntimeException()`

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+      this.requestValidator.validateOrThrow(requestContext, request);
+      String ruleId = request.getId();
+      LabelApplicationRule labelApplicationRule =
+          LabelApplicationRule.newBuilder().setId(ruleId).setData(request.getData()).build();
+      if (this.labelApplicationRuleStore.getObject(requestContext, ruleId).isEmpty()) {
+        responseObserver.onError(new StatusRuntimeException(Status.NOT_FOUND));
+        return;
+      }
+      LabelApplicationRule updateLabelApplicationRule =
+          this.labelApplicationRuleStore.upsertObject(requestContext, labelApplicationRule);
```
aaron-steinfeld, line 131: Suggestion (avoids extra return and removes assumption about structure of a rule):

```java
      LabelApplicationRule existingRule =
          this.labelApplicationRuleStore
              .getObject(requestContext, request.getId())
              .orElseThrow(Status.NOT_FOUND::asException);

      LabelApplicationRule updateLabelApplicationRule =
          existingRule.toBuilder().setData(request.getData()).build();
```

What I mean by the assumption is that if existingRule contained other fields that weren't part of the update flow (imagine something like creation timestamp), they'll be preserved, which would need to be explicit when creating a brand new rule object from scratch in the current approach.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -0,0 +1,162 @@
...
+
+  @Override
+  public void deleteLabelApplicationRule(
+      DeleteLabelApplicationRuleRequest request,
+      StreamObserver<DeleteLabelApplicationRuleResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      this.requestValidator.validateOrThrow(requestContext, request);
+      String ruleId = request.getId();
+      if (this.labelApplicationRuleStore.getObject(requestContext, ruleId).isEmpty()) {
```
aaron-steinfeld, line 150: this check already happens lower down, doing it here is a redundant fetch.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
kamaleshnneerasa: Yes, you are right, but I think we still need to support it. I don't think we can validate a oneOf field using the given functions
Take this example 
```proto
message WrapperMessage {
  InnerMessage inner_message = 1;
 
  message InnerMessage {
        oneOf InnerMessage{
             string str_value = 1;
             int int_value = 2;
             ...
        }
  }
}
```
How do we validate it?

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
kamaleshnneerasa: I don't think its a redundant fetch. Can you point me to the line where this check happens for the second time?

aaron-steinfeld: https://github.com/hypertrace/config-service/blob/318a1e3fd88bbd3fa4cc832d9d4333b6679e3e54/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java#L110-L116

aaron-steinfeld: (by lower down I meant lower in the stack, not in this file 😉 )

kamaleshnneerasa: For some reason I am encountering a NullPointerException when trying to delete a non-existent config.

aaron-steinfeld: Where's the NPE? What's null?

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -146,17 +136,46 @@ public void deleteLabelApplicationRule(
...
+          .orElseThrow(Status.NOT_FOUND::asRuntimeException);
+      this.labelApplicationRuleStore.deleteObject(requestContext, request.getId());
       responseObserver.onNext(DeleteLabelApplicationRuleResponse.getDefaultInstance());
       responseObserver.onCompleted();
     } catch (Exception e) {
       responseObserver.onError(e);
     }
   }
+
+  private static class LabelApplicationRuleStore
```
aaron-steinfeld, line 150: Why choose to make this private static? It's not a problem per se, but more files isn't always a bad thing :)

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
aaron-steinfeld: Discussed offline. This functionality should be added to the mock config service.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
aaron-steinfeld: for one of, we generally want to go in and validate whichever case, right?What my code looks like for that in other cases is for example:


```java

 private void validateEachOneOf(WrapperMessage wrapperMessage) {
    switch (wrapperMessage.getInnerMessageCase()) {
      case STR_VALUE:
        validateNonDefaultPresenceOrThrow(wrapperMessage, WrapperMessage.STR_VALUE_FIELD_NUMBER);
        break;
      case INT_VALUE:
        validateNonDefaultPresenceOrThrow(wrapperMessage, WrapperMessage.INT_VALUE_FIELD_NUMBER);
        break;
// ...
      default:
        throw Status.INVALID_ARGUMENT
            .withDescription("Unexpected case: " + printMessage(wrapperMessage))
            .asRuntimeException();
    }
  }
```

For the primitive cases, we could likely figure out a way to automate it, but I'm not sure we'd want to do implicit validation like that, since a change in proto _should_ correspond to either no change in validation logic (if that were to be a breaking change), or an explicit change (either in a new optional field case, such as one of, or a breaking change).

kamaleshnneerasa: Thanks, this works.

label-application-rule-config-service-impl/src/test/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImplTest.java
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  @Test
+  void getLabelApplicationRule() {
+    Map<String, LabelApplicationRule> createdRulesById = createRules();
+    createdRulesById
+        .keySet()
+        .forEach(
+            id -> {
+              System.out.println(id);
```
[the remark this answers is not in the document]
kamaleshnneerasa: Yes, we can will update.

label-application-rule-config-service-impl/src/test/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImplTest.java
```diff
@@ -0,0 +1,299 @@
...
+    mockGenericConfigService.shutdown();
+  }
+
+  @Test
+  void createLabelApplicationRule() {
+    Map<String, LabelApplicationRule> createdRulesById = createRules();
+    Set<LabelApplicationRuleData> createdData =
+        createdRulesById.values().stream()
+            .map(LabelApplicationRule::getData)
+            .collect(Collectors.toSet());
```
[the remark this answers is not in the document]
kamaleshnneerasa: Yes, will update.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -0,0 +1,193 @@
...
+      default:
+        throwInvalidArgumentException(
+            String.format(
+                "Unexpected Case in %s:%n %s",
+                getName(leafCondition), printMessage(leafCondition)));
+    }
+  }
+
+  private void validateCompositeCondition(CompositeCondition compositeCondition) {
+    if (compositeCondition.getOperator()
```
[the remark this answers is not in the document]
kamaleshnneerasa: Yes, thecommon validation works. Changed the code.

label-application-rule-config-service-impl/src/test/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImplTest.java
```diff
@@ -0,0 +1,299 @@
...
+    assertEquals(
+        expectedRules,
+        response.getLabelApplicationRulesList().stream().collect(Collectors.toUnmodifiableSet()));
+  }
+
+  @Test
+  void updateLabelApplicationRule() {
+    Map<String, LabelApplicationRule> createdRulesById = createRules();
+    LabelApplicationRuleData expectedData = buildSimpleRuleData("auth", "not-valid");
+    String updateRuleId = createdRulesById.keySet().stream().findAny().orElse("default");
```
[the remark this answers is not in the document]
kamaleshnneerasa: Fixed.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
kamaleshnneerasa: I think it is required. Can you ping Kshitiz about this?

saxenakshitiz: we can remove this api and add it if there is a need

PR #62, opened 2021-10-04
fix #310: Changes to publish config change events

saxenakshitiz:


config-service-change-event-api/src/main/proto/org/hypertrace/config/change/event/v1/config_change_event_value.proto
```diff
@@ -0,0 +1,28 @@
...
+message ConfigChangeEventValue {
+  oneof event {
+    ConfigCreateEvent create_event = 1;
+    ConfigUpdateEvent update_event = 2;
+    ConfigDeleteEvent delete_event = 3;
+  }
+}
+
+message ConfigCreateEvent {
+  google.protobuf.Value created_config = 1;
```
aaron-steinfeld, line 18: So our options here that I see are basically:
1. Use the actual types, and have a specific event per config type (doesn't seem feasible)
2. Use struct (which we know can represent all of our configs, since we use it that way internally today)
3. Use a byte and/or json string.

So given the last two options, what benefits do you see to the struct vs byte/json? drawbacks?

config-service-change-event-generator/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
+plugins {
+  `java-library`
+  jacoco
+  id("org.hypertrace.jacoco-report-plugin")
+}
+
+dependencies {
+  api(project(":config-service-change-event-api"))
+  api("com.typesafe:config:1.4.1")
```
aaron-steinfeld, line 9: fix libs to use version catalog

config-service-change-event-generator/src/main/java/org/hypertrace/config/service/change/event/api/ConfigChangeEventGenerator.java
```diff
@@ -0,0 +1,49 @@
...
+   * Send create notification for newly added config
+   *
+   * @param tenantId tenant Id
+   * @param resourceName resource name
+   * @param resourceNamespace resource namespace
+   * @param context context
+   * @param config newly created config
+   */
+  void sendCreateNotification(
+      String tenantId, String resourceName, String resourceNamespace, String context, Value config);
```
aaron-steinfeld, line 18: Configs may or may not have a context, we should be sure to account for that so we don't have to pass special strings and/or nulls across this API.

aaron-steinfeld: Coming back to this after, I think this is picking up the special default context string because it's pulling in the docs so far down the stack - we don't want to be publishing that.

saxenakshitiz:
Addresses https://github.com/hypertrace/hypertrace/issues/310

config-service-change-event-api/build.gradle.kts
```diff
@@ -0,0 +1,26 @@
...
+
+plugins {
+  `java-library`
+  id("com.google.protobuf") version "0.8.17"
+  id("org.hypertrace.publish-plugin")
+}
+
+protobuf {
+  protoc {
+    artifact = "com.google.protobuf:protoc:3.18.0"
```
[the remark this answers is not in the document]
saxenakshitiz: Done

config-service-change-event-generator/build.gradle.kts
saxenakshitiz: Done

config-service-change-event-api/src/main/java/org/hypertrace/config/change/event/v1/ConfigChangeEventKey.java
```diff
@@ -0,0 +1,41 @@
+package org.hypertrace.config.change.event.v1;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Objects;
+
+/** Key for config change event */
+public class ConfigChangeEventKey {
```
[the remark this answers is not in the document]
saxenakshitiz: Yes, the consumer can determine from the key which config is being updated. They also need this information to deserialize.

config-service-change-event-generator/src/main/java/org/hypertrace/config/service/change/event/api/ConfigChangeEventGenerator.java
saxenakshitiz: if context is default context, then it will not be published. Otherwise it will be published?

saxenakshitiz: We will need to pass null for context, if it is default context. Isn't it?

config-service-change-event-api/src/main/proto/org/hypertrace/config/change/event/v1/config_change_event_value.proto
saxenakshitiz: Converting to byte/json is extra step for serialisation as well as deserialisation. That is the reason I went with option 2.

config-service-change-event-api/src/main/java/org/hypertrace/config/change/event/v1/ConfigChangeEventKey.java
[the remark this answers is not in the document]
saxenakshitiz: Since you pointed out issue with using protobuf for key, I went with json for keys.

config-service-change-event-generator/src/main/java/org/hypertrace/config/service/change/event/api/ConfigChangeEventGenerator.java
saxenakshitiz: Or you mean we don't need to send context at all?

config-service-change-event-api/src/main/java/org/hypertrace/config/change/event/v1/ConfigChangeEventKey.java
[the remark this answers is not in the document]
saxenakshitiz: One point I overlooked in the page you had shared -
```
The indeterminism is caused by -
1. the absence of guarantees on how known or unknown fields should be ordered upon writing them
2. map serialisation relies on iteration, and there is no ordering guarantee while iterating over map items
```
A simple proto with string values will never run in to above issue. Based on this I am pretty confident in exposing key as proto itself.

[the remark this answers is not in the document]
aaron-steinfeld: > A simple proto with string values will never run in to above issue. Based on this I am pretty confident in exposing key as proto itself.

"how known fields should be ordered" would be the issue that could apply here, since we're (at the moment) sending all fields and not using maps. That still leaves risk in evolution though, if a new field were to be added.

But assuming it works, the bigger risk we should consider IMO - will the next person realize these considerations when they reference this to set up a new topic with proto keys? We just want to be careful when establishing a pattern.

So, a summary of my thoughts:
- The point about bringing up proto concerns for keys was to consider and mitigate any issues, but doesn't mean we shouldn't use them necessarily
- Using proto gives us multi-language support, backwards compatibility/breaking change detection support and is our standard elsewhere, so should be our default unless we have serious reasons not to use it
- JSON as a transport format doesn't give us much advantage, as we could just as easily represent a proto key in json. It's only a question of getting a consistent serializer, or determining that the off the shelf proto serializer is good enough for our needs 
- For the vast majority of our usages, we have one publisher and N subscribers (more often than not one, but sometimes more). At least, I can't think of any topics today where we use multiple publishers to one topic, although it's a perfectly valid topology. The concern about key consistency is on the serialization side, which seems at least partially mitigated by using a single publisher unless the serializer is self-inconsistent.
- The same thought process around having a more complex key seems like it goes for the entity change topic too - the consumer should be able to read the type off the key before requiring deserializing the values. That is, give me only API changes for example - and we don't want magic concatenated strings if we need to read them.

config-service-change-event-generator/src/main/java/org/hypertrace/config/service/change/event/api/ConfigChangeEventGenerator.java
aaron-steinfeld: `public static final String DEFAULT_CONTEXT = "DEFAULT-CONTEXT";`

So this is a perfectly valid style of config, that we just internally use this constant to denote. It doesn't have an ID, so would generally be tenant-wide. When we publish changes in a config like that, we don't want to expose that internal string - instead, we'd ideally make the context field on `ConfigChangeEventKey` optional.

SrikarMannepalli:
Is it possible to have the initiator name and id also as part of the messages ?

SrikarMannepalli:
> > > > > > Is it possible to have the initiator name and id also as part of the messages ?
> > > > > 
> > > > > 
> > > > > What do you mean by initiator? I am making changes to publish the config class name as part of the message. You will get details like - `org.hypertrace.label.config.service.v1.Label`. Is this enough?
> > > > > Config doesn't have id. May have optional context. That will be available.
> > > > 
> > > > 
> > > > I think he is referring to the user who initiated the change
> > > 
> > > 
> > > Yes. I was referring to the user who has initiated the change.
> > 
> > 
> > As of now, user Id is not set in config. It is still a to-do. Not sure we will always have the user Id, and the config is not always created via user action. For example, labels can be created dynamically based on a rule.
> 
> Can we get it from request context where available ? @SrikarMannepalli which events require the initiator ?

I think we need the initiator for all of the user related events.(All of the Security configuration changes, sensitive data changes and product access changes)

aaron-steinfeld (commented):
Whoops, had this comment stuck in queue. WRT to the initiator, yes we should be able to do this here and on the entity change events, add a field to change event populated via `RequestContext.getUserId()` (will be optional and missing for system-generated changes).

helm/values.yaml
```diff
@@ -91,3 +92,20 @@ logConfig:
...
+  enabled: true
+  jobName: config-service-kafka-topics-creator
+  helmHook: pre-install,pre-upgrade
+  kafka:
+    topics:
+      config-change-events:
+        replicationFactor: 3
+        partitions: 8
+        configs:
+          retention.bytes: 1073741824 # default = -1
```
jcchavezs, line 106: curious about what is this value (in MB?) and also why this is the right value and not for example the half of it. It is good to have this explanation in the code so next person willing to change it has the right context.

SrikarMannepalli:
> > > > > > Whoops, had this comment stuck in queue. WRT to the initiator, yes we should be able to do this here and on the entity change events, add a field to change event populated via `RequestContext.getUserId()` (will be optional and missing for system-generated changes).
> > > > > 
> > > > > 
> > > > > Can we include name or email as well ?
> > > > 
> > > > 
> > > > We could - all three are on the request context - but we could also look up that information from the ID, so just a question of how much we want to flatten. My preference is to keep it lighter.
> > > > Do we need a user context to retrieve userInfo like first name, last name etc ?
> > > 
> > > 
> > > @SrikarMannepalli what attributes of the user are needed ?
> > 
> > 
> > We want to display the first name and last name in the messages.
> 
> @surajpuvvada @SrikarMannepalli Do we need to provide all this information as part of kafka message or they can be retrieved using user-id?

I think we need user context to retrieve some of the user info like first name and last name. We don't store it on our side. So I think this has to be provided as part of Kafka message.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -90,6 +93,9 @@ message ContextSpecificConfig {
 
   // timestamp (in milliseconds since epoch) at which this config was last updated
   int64 update_timestamp = 4;
+
+  // prev config value which applies to the above context
+  optional google.protobuf.Value prevConfig = 5;
```
aaron-steinfeld, line 98: naming should be snake_case, but why is this needed here? Only makes sense on upsert response, right?

saxenakshitiz: Yeah. Already fixed it.

config-service-change-event-api/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+    java {
+      srcDirs("build/generated/source/proto/main/java")
+    }
+  }
+}
+
+dependencies {
+  api(libs.protobuf.java)
+
+  api(libs.jackson.databind)
```
aaron-steinfeld, line 27: jackson shouldn't be a dependency of the API

config-service-change-event-api/src/main/proto/org/hypertrace/config/change/event/v1/config_change_event_key.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+option java_multiple_files = true;
+
+package org.hypertrace.config.change.event.v1;
+
+message ConfigChangeEventKey {
+  string tenant_id = 1;
+  string config_type = 2;
```
aaron-steinfeld, line 9: Does it make these harder to use having to work with both keys and values on the consumer side? Seems like it could be a bit clunky, but not sure. I guess the alternative is to dupe all 3 into the value, which seems bad too.

config-service-change-event-generator/src/main/java/org/hypertrace/config/service/change/event/api/ConfigChangeEventGenerator.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * Send create notification for newly added config
+   *
+   * @param requestContext requestContext
+   * @param configType config type
+   * @param context context
+   * @param config newly created config
+   */
+  void sendCreateNotification(
```
aaron-steinfeld, line 17: can we create overloads without a context field, since that's optional for configs (better than explicitly passing nulls, which is I'm guessing what's required right now)

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -47,6 +47,7 @@ public void upsertConfig(
               .setConfig(request.getConfig())
               .setCreationTimestamp(contextSpecificConfig.getCreationTimestamp())
               .setUpdateTimestamp(contextSpecificConfig.getUpdateTimestamp())
+              .setPrevConfig(contextSpecificConfig.getPrevConfig())
```
aaron-steinfeld, line 50: this should be optionally set

config-service/build.gradle.kts
```diff
@@ -75,6 +76,7 @@ dependencies {
     runtimeOnly("io.netty:netty-codec-http2:4.1.68.Final")
     runtimeOnly("io.netty:netty-handler-proxy:4.1.68.Final")
   }
+  runtimeOnly(libs.kafka.protobuf.serializer)
```
aaron-steinfeld, line 79: would it make sense to specify this in the generator? Seems like this leaks a concern from the generator up here.

config-service/src/main/java/org/hypertrace/config/service/ConfigServicesFactory.java
```diff
@@ -30,14 +32,17 @@
...
+
     return List.of(
         new ConfigServiceGrpcImpl(configStore),
         new SpacesConfigServiceImpl(configChannel),
         new LabelsConfigServiceImpl(configChannel, config),
-        new LabelApplicationRuleConfigServiceImpl(configChannel),
-        new EventConditionConfigServiceImpl(configChannel),
-        new NotificationRuleConfigServiceImpl(configChannel),
-        new NotificationChannelConfigServiceImpl(configChannel));
+        new LabelApplicationRuleConfigServiceImpl(configChannel, configChangeEventGenerator),
```
aaron-steinfeld, line 42: Not sure if any of these are using guice (assuming not due to the explicit construction), but exposing a guice module for the generator would make adding it to guice-enabled services less invasive.

object-store/src/main/java/org/hypertrace/config/objectstore/DefaultObjectStore.java
```diff
@@ -19,14 +21,17 @@
...
   private final String resourceNamespace;
   private final String resourceName;
+  private final ConfigChangeEventGenerator configChangeEventGenerator;
 
   protected DefaultObjectStore(
       ConfigServiceBlockingStub configServiceBlockingStub,
       String resourceNamespace,
-      String resourceName) {
+      String resourceName,
+      ConfigChangeEventGenerator configChangeEventGenerator) {
```
aaron-steinfeld, line 30: Could we overload this to make it optional? That way, implementers can opt out (and we save ourselves from this being a breaking change)

object-store/src/main/java/org/hypertrace/config/objectstore/DefaultObjectStore.java
```diff
@@ -56,20 +61,32 @@ protected DefaultObjectStore(
...
-        .orElseThrow(Status.INTERNAL::asRuntimeException);
+    T upsertedObject =
+        this.buildObjectFromValue(response.getConfig())
+            .orElseThrow(Status.INTERNAL::asRuntimeException);
+
+    if (response.hasPrevConfig()) {
+      configChangeEventGenerator.sendUpdateNotification(
+          context,
+          upsertedObject.getClass().getName(),
+          null,
```
aaron-steinfeld, line 82: here are those nulls 😉

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -180,4 +180,34 @@ private ContextSpecificConfig getContextSpecificConfig(ConfigDocument configDocu
...
+          .setConfig(configDocument.getConfig())
+          .setContext(configDocument.getContext())
+          .setCreationTimestamp(configDocument.getCreationTimestamp())
+          .setUpdateTimestamp(configDocument.getUpdateTimestamp())
+          .setPrevConfig(existingConfig.getConfig())
+          .build();
+    }
+  }
+
+  private ContextSpecificConfig getContextSpecificConfig1(
```
aaron-steinfeld, line 199: `getContextSpecificConfig1` 🤨

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -180,4 +180,34 @@ private ContextSpecificConfig getContextSpecificConfig(ConfigDocument configDocu
...
+      ConfigDocument configDocument, ContextSpecificConfig existingConfig) {
+    if (ConfigServiceUtils.isNull(existingConfig.getConfig())) {
+      return getContextSpecificConfig(configDocument);
+    } else {
+      return ContextSpecificConfig.newBuilder()
+          .setConfig(configDocument.getConfig())
+          .setContext(configDocument.getContext())
+          .setCreationTimestamp(configDocument.getCreationTimestamp())
+          .setUpdateTimestamp(configDocument.getUpdateTimestamp())
+          .setPrevConfig(existingConfig.getConfig())
```
aaron-steinfeld, line 194: Ah - so this is why this got added. We should find another way to return this such as a wrapper, because this field doesn't make sense on the api, only for internal usage.

config-service-change-event-api/src/main/proto/org/hypertrace/config/change/event/v1/config_change_event_key.proto
saxenakshitiz: Should be fine in my view. No need to duplicate information in two places

config-service-change-event-generator/src/main/java/org/hypertrace/config/service/change/event/api/ConfigChangeEventGenerator.java
saxenakshitiz: Done

config-service-change-event-api/build.gradle.kts
saxenakshitiz: Removed

config-service/build.gradle.kts
saxenakshitiz: Moved to generator

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
saxenakshitiz: Done

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -93,6 +93,20 @@ message ContextSpecificConfig {
 
   // timestamp (in milliseconds since epoch) at which this config was last updated
   int64 update_timestamp = 4;
+}
+
+message InternalContextSpecificConfig {
```
aaron-steinfeld, line 98: This doesn't need to be a proto, and it doesn't belong to the API. Since it's internal, we can just build a POJO containing a ContextSpecificConfig and an optional prev config.

saxenakshitiz: Done

config-service-change-event-api/src/main/proto/org/hypertrace/config/change/event/v1/config_change_event_value.proto
```diff
@@ -0,0 +1,30 @@
...
+    ConfigCreateEvent create_event = 1;
+    ConfigUpdateEvent update_event = 2;
+    ConfigDeleteEvent delete_event = 3;
+  }
+  optional string user_id = 4;
+  optional string user_name = 5;
+}
+
+message ConfigCreateEvent {
+  string created_config = 1;
```
aaron-steinfeld, line 20: since we're storing json in this string, can we clarify the name to `created_config_json` (and below)

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -180,4 +181,15 @@ private ContextSpecificConfig getContextSpecificConfig(ConfigDocument configDocu
...
         .build();
   }
+
+  private InternalContextSpecificConfig getInternalContextSpecificConfig(
+      ConfigDocument configDocument, ContextSpecificConfig existingConfig) {
+    if (ConfigServiceUtils.isNull(existingConfig.getConfig())) {
+      return new InternalContextSpecificConfig(
+          getContextSpecificConfig(configDocument), Optional.empty());
+    } else {
+      return new InternalContextSpecificConfig(
```
aaron-steinfeld, line 191: nit- cleaner to overload constructor rather than accept an optional (optional params could technically be null, which makes it better suited as a return type than a param type).

config-service-impl/src/main/java/org/hypertrace/config/service/store/InternalContextSpecificConfig.java
```diff
@@ -1,11 +1,29 @@
...
+    this.prevConfigOptional = Optional.empty();
+  }
+
+  public InternalContextSpecificConfig(
+      ContextSpecificConfig contextSpecificConfig, com.google.protobuf.Value prevConfig) {
+    this.contextSpecificConfig = contextSpecificConfig;
+    this.prevConfigOptional = Optional.of(prevConfig);
+  }
+
+  public ContextSpecificConfig getContextSpecificConfig() {
```
aaron-steinfeld, line 22: You could still leave this as `@Value` with your own constructors, up to you (besides the getters, it will give you equals/hashcode/tostring - important if you're using these anywhere in any of those ways).

aaron-steinfeld (approved):

PR #64, opened 2021-10-06
feat: return deleted value from object store's delete

aaron-steinfeld:
## Description
Make ObjectStore's delete return the deleted value, matching the grpc's API.

### Testing
Updated existing unit tests, verified integration tests

### Checklist:

### DIFF
--- a/alerting-config-service-impl/src/main/java/org/hypertrace/alerting/config/service/EventConditionConfigServiceImpl.java
+++ b/alerting-config-service-impl/src/main/java/org/hypertrace/alerting/config/service/EventConditionConfigServiceImpl.java
@@ -1,6 +1,7 @@
 package org.hypertrace.alerting.config.service;
 
 import io.grpc.Channel;
+import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
 import java.util.UUID;
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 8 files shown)


aaron-steinfeld:
> Deleted APIs should be idempotent. How do we handle that if return value is gone?

So this change isn't touching the delete implementation, it's just changing the client to return the deleted value already returned by the API.

As to the API being idempotent - it is. There is no difference in server state between calling this delete API once or 10 times. What does change is that the first call would return a success response, including that deleted value. Subsequent delete calls would throw a NOT_FOUND status exception instead, but with no effect on server state. Even though this is our own RPC and not rest, these are basically the same semantics expected of REST DELETE which is what I think you're driving at.

alerting-config-service-impl/src/main/java/org/hypertrace/alerting/config/service/EventConditionConfigServiceImpl.java
```diff
@@ -100,7 +101,9 @@ public void deleteEventCondition(
     try {
       RequestContext requestContext = RequestContext.CURRENT.get();
       requestValidator.validateDeleteEventConditionRequest(requestContext, request);
-      eventConditionStore.deleteObject(requestContext, request.getEventConditionId());
+      eventConditionStore
+          .deleteObject(requestContext, request.getEventConditionId())
+          .orElseThrow(Status.NOT_FOUND::asRuntimeException);
```
aaron-steinfeld, line 106: so need to be careful now to go through all usages and change them to throw, since we don't return on delete generally the compiler won't catch the change - this is why I felt throwing in the delete case vs returning the optional was more suited in delete (while get does return, so the optional there felt more natural).

Anyway, we have both versions now so easy to go with either one - what do you think @saxenakshitiz ?

saxenakshitiz: In my view, it is fine to return an optional. It is better that caller make the decision whether to return not_found status or just return nothing.

PR #70, opened 2021-10-13
chore: fix kafka version to match platform

aaron-steinfeld:
## Description
Change kafka version to match other kafka client versions in use throughout platform


gradle/libs.versions.toml
```diff
@@ -5,6 +5,7 @@ hypertrace-grpcutils = "0.5.2"
 hypertrace-framework = "0.1.29"
 lombok = "1.18.20"
 jackson = "2.12.4"
+kafka = "6.0.1"
```
saxenakshitiz, line 8: can we rename it to confluent as it is confluent platform version and not kafka?

aaron-steinfeld: Fixed

PR #72, opened 2021-10-15
feat: add upsert all support

aaron-steinfeld:
## Description
Added support for a new upsertAll api. This can be used for upserting multiple configs in fewer calls. Eventually, we'll optimize its internals to also reduce the roundtrips to the persistence store, but this change encapsulates that tech debt.

### Testing
Added and updated existing UTs

### Checklist:

### DIFF
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
@@ -91,7 +91,10 @@ public class DocumentConfigStore implements ConfigStore {
           previousConfigDoc.flatMap(this::convertToContextSpecificConfig);
       long updateTimestamp = clock.millis();
       long creationTimestamp =
-          previousConfigDoc.map(ConfigDocument::getCreationTimestamp).orElse(updateTimestamp);
+          previousConfigDoc
+              .filter(configDocument -> !ConfigServiceUtils.isNull(configDocument.getConfig()))
+              .map(ConfigDocument::getCreationTimestamp)
+              .orElse(updateTimestamp);
       long newVersion =
           previousConfigDoc
               .map(ConfigDocument::getConfigVersion)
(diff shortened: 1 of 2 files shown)


config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
```diff
@@ -23,34 +26,42 @@
...
+
+  /**
+   * Write each of the provided config value associated with the specified config resource to the
+   * store.
+   *
+   * @param resourceContextValueMap
+   * @param userId
+   * @return the upserted configs
+   */
+  List<UpsertedConfig> writeAllConfigs(
```
surajpuvvada, line 63: what is the error handling behavior ?

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -64,68 +80,69 @@ private Datastore initDataStore(Config config) {
...
     }
   }
 
   @Override
-  public ContextSpecificConfig getConfig(ConfigResource configResource) throws IOException {
-    return getConfig(configResource, getLatestVersion(configResource));
+  public List<UpsertedConfig> writeAllConfigs(
+      Map<ConfigResourceContext, Value> resourceContextValueMap, String userId) throws IOException {
+    // TODO keep pushing this down into doc store
+    List<UpsertedConfig> list = new ArrayList<>(resourceContextValueMap.size());
```
surajpuvvada, line 126: so would docstore be able to leverage bulk upsert APIs ?

aaron-steinfeld: That's the idea - just split the PR. The current PR handles all the API changes, which is a fairly significant refactor, but doesn't really change the implementation in any meaningful way. A future PR needs only swap out this implementation to use the doc store's bulk upsert.

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
aaron-steinfeld: Currently, it's identical to calling N write operations sequentially. If a later one fails, it aborts at that point leaving the original (this is the same behavior in the current abstraction one layer up in ObjectStore). Ideally, with a doc store API change we can make this transactional with all or nothing semantics (this came up with Entity Service in conversations with @saxenakshitiz too, but I believe as of today doc store itself doesn't support this - another reason I paused this change where I did).

PR #82, opened 2021-11-22
Changes to support bulk create of labels

saxenakshitiz:


labels-config-service-api/src/main/proto/org/hypertrace/label/config/service/v1/labels_config_service.proto
```diff
@@ -11,6 +11,9 @@ service LabelsConfigService {
   // create label
   rpc CreateLabel(CreateLabelRequest) returns (CreateLabelResponse) {}
 
+  // create labels
+  rpc BulkCreateLabels(BulkCreateLabelsRequest) returns (BulkCreateLabelsResponse) {}
```
aaron-steinfeld, line 15: nit: bulk not needed here
```suggestion
  rpc CreateLabels(CreateLabelsRequest) returns (CreateLabelsResponse) {}
```

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -136,6 +139,59 @@ public void createLabel(
     }
   }
 
+  @Override
+  public void bulkCreateLabels(
+      BulkCreateLabelsRequest bulkCreateLabelsRequest,
+      StreamObserver<BulkCreateLabelsResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      Lock labelsLock = this.stripedLabelsLock.get(requestContext.getTenantId());
```
aaron-steinfeld, line 148: :( We've got a fairly high priority ticket to get rid of this locking - I suppose this is the same as the single create label lock and will be removed at the same time?

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -136,6 +139,59 @@ public void createLabel(
...
+        responseObserver.onError(new StatusRuntimeException(Status.ABORTED));
+      }
+    } catch (Exception e) {
+      responseObserver.onError(e);
+    }
+  }
+
+  private Optional<Label> buildLabelFromRequest(BulkCreateLabelsRequest.LabelRequest request) {
+    LabelData labelData = request.getData();
+    if (systemLabelsKeyLabelMap.containsKey(labelData.getKey())
```
aaron-steinfeld, line 181: I'm tempted to error if they try to create something that already exists - at least on system labels, where there's no worry on races. On user labels, we may still end up in an inconsistent state if we don't error this, since the caller may thing it was created with the parameter they specified and it was actually created with some other parameters. This is particularly problematic if we allow the caller to pre-define IDs so they don't have to block on the response. Also, would this cause the existing label to be omitted from the response? that's problematic too - the caller wouldn't be able to get IDs back

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -136,6 +139,59 @@ public void createLabel(
...
+        || isDuplicateKey(labelData.getKey())) {
```
aaron-steinfeld, line 182: isDuplicateKey does a getAll fetch right? Can we avoid doing that on each request in the bulk create and fetch the exiting ones once?

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -136,6 +139,59 @@ public void createLabel(
...
+      // Creating a label with a name that clashes with one of system labels
+      // name. i
```
aaron-steinfeld, line 184: incomplete comment

labels-config-service-api/src/main/proto/org/hypertrace/label/config/service/v1/labels_config_service.proto
saxenakshitiz: okay

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
saxenakshitiz: Yes,it will be

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
saxenakshitiz: I had ignored duplicate labels for the case when two enricher instance is trying to create duplicate labels as part of bulk create. Handling error will be difficult in that case. We will have bail out without creating labels.

aaron-steinfeld: Yeah, handling that case is important (although what's the partitioning like? If it's by tenant then that would greatly mitigate this concern, right).

So I think we need to think about the pre-defining IDs case - even if we don't do it yet - because that may influence how this gets built. If two clients are asking to create the same label, what does that ask look like and what does the response look like for the one that arrives first and the one that arrives second?

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
saxenakshitiz: Fixed this part

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
saxenakshitiz: Fixed it

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
saxenakshitiz: Discussed this offline and conclusion is that for existing labels, we will still return them as part of response.

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -136,6 +137,66 @@ public void createLabel(
...
+                          labelRequest, existingLabelsMap, newLabels, allLabels));
+          if (!newLabels.isEmpty()) {
+            List<Label> createdLabels =
+                labelStore.upsertObjects(requestContext, newLabels).stream()
+                    .map(org.hypertrace.config.objectstore.ConfigObject::getData)
+                    .collect(Collectors.toList());
+            allLabels.addAll(createdLabels);
+          }
+          responseObserver.onNext(
+              GetOrCreateLabelsResponse.newBuilder().addAllLabels(allLabels).build());
```
aaron-steinfeld, line 165: nit - can we preserve the order of labels on the response to match that of the request? That'd be my assumption as a caller

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -242,14 +303,13 @@ public void deleteLabel(
   }
 
   private boolean isDuplicateKey(String key) {
+    return getLabelsMap().containsKey(key);
+  }
+
+  private Map<String, Label> getLabelsMap() {
```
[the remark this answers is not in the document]
saxenakshitiz: Done

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -136,6 +137,66 @@ public void createLabel(
...
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      Lock labelsLock = this.stripedLabelsLock.get(requestContext.getTenantId());
+      if (labelsLock.tryLock(WAIT_TIME.getSeconds(), TimeUnit.SECONDS)) {
+        try {
+          final Map<String, Label> existingLabelsMap = getLabelsMap();
+          final List<Label> newLabels = new ArrayList<>();
+          final List<Label> allLabels = new ArrayList<>();
+          request.getRequestsList().stream()
+              .forEach(
```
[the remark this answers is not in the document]
saxenakshitiz: Done

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
saxenakshitiz: okay. Made this change.

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -136,6 +137,66 @@ public void createLabel(
...
+    }
+  }
+
+  private void buildLabelFromRequest(
+      GetOrCreateLabelsRequest.LabelRequest request,
+      Map<String, Label> existingLabelsMap,
+      List<Label> newLabels,
+      List<Label> allLabels) {
+    LabelData labelData = request.getData();
+    if (systemLabelsKeyLabelMap.containsKey(labelData.getKey())) {
```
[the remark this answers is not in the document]
saxenakshitiz: Done

aaron-steinfeld (approved):

PR #85, opened 2021-12-03
Move max dynamic rules config under label application rule config

saxenakshitiz:


helm/values.yaml
```diff
@@ -83,7 +83,6 @@ configServiceConfig:
   mongo:
     host: mongo
     url: ""
-  maxDynamicLabelApplicationRulesPerTenant: 100
```
pavan-traceable, line 86: Now where is this defined in configmap then ?

pavan-traceable (approved):

PR #86, opened 2021-12-06
added new object store to use external context

SrikarMannepalli:
## Description
Added a new object store abstraction to use context that has been provided separately. (Needed in cases where the identifying context cannot be extracted either from the data object or request context)

### DIFF
--- a/object-store/src/main/java/org/hypertrace/config/objectstore/ExternalContextIdentifiedObjectStore.java
+++ b/object-store/src/main/java/org/hypertrace/config/objectstore/ExternalContextIdentifiedObjectStore.java
@@ -1,0 +1,157 @@
+package org.hypertrace.config.objectstore;
+
+import com.google.protobuf.Value;
+import io.grpc.Status;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.config.service.change.event.api.ConfigChangeEventGenerator;
+import org.hypertrace.config.service.v1.ConfigServiceGrpc.ConfigServiceBlockingStub;
+import org.hypertrace.config.service.v1.ContextSpecificConfig;
+import org.hypertrace.config.service.v1.DeleteConfigRequest;
+import org.hypertrace.config.service.v1.GetConfigRequest;
+import org.hypertrace.config.service.v1.UpsertConfigRequest;
+import org.hypertrace.config.service.v1.UpsertConfigResponse;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+/**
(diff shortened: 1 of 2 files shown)


object-store/src/main/java/org/hypertrace/config/objectstore/ExternalContextIdentifiedObjectStore.java
```diff
@@ -0,0 +1,157 @@
...
+      return Optional.of(data);
+    } catch (Exception exception) {
+      if (Status.fromThrowable(exception).equals(Status.NOT_FOUND)) {
+        return Optional.empty();
+      }
+      throw exception;
+    }
+  }
+
+  public ConfigObject<T> upsertObject(RequestContext context, String dataContext, T data) {
```
aaron-steinfeld, line 85: Be careful with naming. What we call dataContext on upsert is called id on delete.

object-store/src/main/java/org/hypertrace/config/objectstore/ExternalContextIdentifiedObjectStore.java
```diff
@@ -0,0 +1,157 @@
...
+
+  protected Value buildValueForChangeEvent(T data) {
+    return this.buildValueFromData(data);
+  }
+
+  protected String buildClassNameForChangeEvent(T data) {
+    return data.getClass().getName();
+  }
+
+  public Optional<T> getData(RequestContext context) {
```
aaron-steinfeld, line 63: get data should take any number of context/ids.

aaron-steinfeld (commented):
getAll, upsertAll are missing. This is basically a more generalized version of IdentifiedObjectStore (where `getContextFromData` is provided on request rather than derived) so a useful exercise would actually be to ensure you can implement IdentifiedObjectStore using this new class.

SrikarMannepalli:
@aaron-steinfeld Would it be good idea to just overload the upsert and get methods in Identified Object store rather than have a new object store?

aaron-steinfeld:
> @aaron-steinfeld Would it be good idea to just overload the upsert and get methods in Identified Object store rather than have a new object store?

The API of identified object store doesn't make sense for this use case (e.g. upsert requires both an object and id here, while for IdentifiedObjectStore just an object is sufficient), so rather than overloading them or using any inheritance I'd suggest creating the new more generic object store class, then use that as the implementation of IdentifiedObjectStore similar to what we did with Contextually Identified Object Store. The latter part isn't critical, but I think it's a good long term decision because it:
- Confirms you're implementing all of the APIs the other stores use
- Allows you to confirm the behavior of the new store because it'll be used by IdentifiedObjectStore's tests
- Reduces code duplication

PR #96, opened 2022-02-23
update exclude span rules apis to expose creation and last updated timestamp

SrikarMannepalli:


span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/utils/TimestampConverter.java
```diff
@@ -0,0 +1,13 @@
+package org.hypertrace.span.processing.config.service.utils;
+
+import com.google.protobuf.Timestamp;
+import java.time.Instant;
+
+public class TimestampConverter {
+  public static Timestamp convert(Instant instant) {
```
skjindal93, line 7: We should not have static methods. It makes it hard to unit test. You can inject the `TimestampConverter` wherever needed

skjindal93 (approved):

PR #97, opened 2022-02-25
add enable/disable support for exclude span rules

SrikarMannepalli:


span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto
```diff
@@ -66,8 +67,9 @@ message ExcludeSpanRuleMetadata {
 
 message UpdateExcludeSpanRule {
   string id = 1;
-  string name = 2;
+  optional string name = 2;
```
skjindal93, line 70: Why optionals? Clients can send the same values in the update messages (_if they really want to keep them as same values_). And, they would be updated as is. We shouldn't really be caring about the presence of these values

skjindal93 (approved):

PR #99, opened 2022-02-28
add impl

SrikarMannepalli:


span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/store/ApiNamingRulesConfigStore.java
```diff
@@ -0,0 +1,73 @@
...
+import org.hypertrace.core.grpcutils.context.RequestContext;
+import org.hypertrace.span.processing.config.service.utils.TimestampConverter;
+import org.hypertrace.span.processing.config.service.v1.ApiNamingRule;
+import org.hypertrace.span.processing.config.service.v1.ApiNamingRuleDetails;
+import org.hypertrace.span.processing.config.service.v1.ApiNamingRuleMetadata;
+
+public class ApiNamingRulesConfigStore extends IdentifiedObjectStore<ApiNamingRule> {
+
+  private static final String API_NAMING_RULES_RESOURCE_NAME = "api-naming-rules";
+  private static final String API_NAMING_RULES_CONFIG_RESOURCE_NAMESPACE =
```
skjindal93, line 21: We should keep the same namespace, we used for exclude rules. We should also change the namespace name for exclude rules config from `exclude-span-rules-config` to `span-processing-rules-config`

SrikarMannepalli: Done.

skjindal93 (approved):

PR #104, opened 2022-03-10
Adding serde for config change event key and value

saxenakshitiz:


laxmanchekka (approved):
looking good. If possible, Serializer and Deserializer implement them together as a pair in a same outer class. Like JaegerSpanSerde.
makes it compact and SerDe are used together as pair hence finding them makes it easier





[12:16](https://traceableai.slack.com/archives/D022WE7G98V/p1646894774332749)
just a suggestion. proceed with merging if you want to proceed with current implementation

saxenakshitiz:
> looking good. If possible, Serializer and Deserializer implement them together as a pair in a same outer class. Like JaegerSpanSerde. makes it compact and SerDe are used together as pair hence finding them makes it easier
> 
> just a suggestion. proceed with merging if you want to proceed with current implementation

event producer config requires only key and value serializer and keeping those classes as inner classes in Serde will add $ to generated classname. Using class names with $ in it didn't look right in application.conf

PR #107, opened 2022-03-14
Removing version from document store key

saxenakshitiz:
Removing version from document store key. Using static "latest" instead. This will ensure only latest version is maintained in the document store. The version history is published by config change events, and can be potentially captured in audit logs.

### DIFF
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocumentKey.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocumentKey.java
@@ -9,9 +9,9 @@ import org.hypertrace.core.documentstore.Key;
 public class ConfigDocumentKey implements Key {
 
   private static final String SEPARATOR = ":";
+  private static final String LATEST_VERSION = "latest";
 
   ConfigResourceContext configResourceContext;
-  long configVersion;
 
   @Override
   public String toString() {
@@ -21,6 +21,6 @@ public class ConfigDocumentKey implements Key {
         configResourceContext.getConfigResource().getResourceNamespace(),
         configResourceContext.getConfigResource().getTenantId(),
         configResourceContext.getContext(),
-        String.valueOf(configVersion));
+        LATEST_VERSION);
   }
 }
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
@@ -7,9 +7,6 @@ import static org.hypertrace.config.service.store.ConfigDocument.RESOURCE_NAMESP
 import static org.hypertrace.config.service.store.ConfigDocument.TENANT_ID_FIELD_NAME;
 import static org.hypertrace.config.service.store.ConfigDocument.VERSION_FIELD_NAME;
 
-import com.google.common.cache.CacheBuilder;
-import com.google.common.cache.CacheLoader;
-import com.google.common.cache.LoadingCache;
 import com.google.protobuf.Value;
 import com.typesafe.config.Config;
 import java.io.IOException;
@@ -23,7 +20,6 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Optional;
 import java.util.Set;
-import java.util.concurrent.TimeUnit;
 import lombok.extern.slf4j.Slf4j;
 import org.hypertrace.config.service.ConfigResource;
 import org.hypertrace.config.service.ConfigResourceContext;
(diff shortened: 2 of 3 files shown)


config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocumentKey.java
```diff
@@ -9,9 +9,9 @@
 public class ConfigDocumentKey implements Key {
 
   private static final String SEPARATOR = ":";
+  private static final String LATEST_VERSION = "latest";
```
avinashkolluru, line 12: How is this backward compatible? There would be existing docs which have `:<n>` in the id.

saxenakshitiz: The lookup in config service is based on tenantId, resource namespace, resource name and optionally context. The results are retrieved using that, sorted in descending order by config version. In case of latest version, the result is limited to 1. So version in key is just to maintain history in mongo collection. Once you generate key independent of version, you will have only latest version of config document in mongo collection.

saxenakshitiz: BTW I have already validated this approach by running traceable config service.
I had pre-created few data types with older version of traceable config service, updated it few times, restarted traceable config service with suggested change and then updated data types few more times. It works as expected without issues.

avinashkolluru: Did you try with postgres too?

saxenakshitiz: > So the order by on version field is working fine when there is a mix of numbers and text?
Version is still a number. Just that version is not used in the key. Version is replaced with static `latest`.

aaron-steinfeld: > Version is still a number. Just that version is not used in the key. Version is replaced with static latest.

So all this would solve is the retention problem, and only for newly created docs (e.g. if I had 10 docs when this change landed, I would create one more then use that). I see three separate problems that we're trying to solve:
1. The unbounded growth of versions
2. GetAll being very memory and/or compute intensive due to 1 (because all versions must be fetched to resolve them)
3. Write being a multi-step, transactional operation to support incrementing the version

I'm not really concerned about the newly-created bit since that can be handled offline, but we lose some of the main benefits of this change if we keep versioning the written doc. Namely, we still have to do multiple queries to write, and thus use locks to make it transactional. Removing those locks is, IMO, the biggest goal here. Those are what prevents this service from scaling.

So maybe this PR is the first on that path, but let's try to think through how we'd get there (whether it's in one go or phases) before merging this.

saxenakshitiz: > The unbounded growth of versions

This will be fixed by this PR + cron job added to remove older versions. We can tune cron job to just maintain a single latest version

> GetAll being very memory and/or compute intensive due to 1 (because all versions must be fetched to resolve them)

As you said, fixing above issue should fix this as well.

> Write being a multi-step, transactional operation to support incrementing the version

This problem is entirely in a different league. Whether we version or not, we need to support concurrent updates of the config. We have discussed distributed locking to achieve that in the past. Either we go that route, or we can use logic used in atomic implementation in java to compare and set(use version for this purpose).

aaron-steinfeld: > This problem is entirely in a different league. Whether we version or not, we need to support concurrent updates of the config.

I guess this is worth its own discussion, because I see this as the same problem. I'd call concurrent updates a nice to have, but it's not something we have elsewhere in the platform so I don't think we need to go to great lengths to start adding it here. If anything, this would be the last place to get it since those other places, like entity service for example, are written at high frequency during ingestion, while config service is being updated at human timescale.

The distributed locking discussions were based on the current versioning mechanism - that's what I'm trying to avoid.

aaron-steinfeld: > However it requires version of config to be propagated in all responses, and send back as part of config updates.

This is the generalized overwrite problem, and I think it's most performantly addressed with optimistic locking - but we don't need version to do that. We also have the update timestamp at our disposal which can be used in the exact same way for example. In either, we'd need to update lots of code to thread enough information through on both and read/write to support. Anyway, I agree (I think this is what you were saying) I would treat that problem as a separate one - this is not something we handle today, either.

> The current locking mechanism does not use the version. It is based on tenant-id, resource namespace, resource name and context(static default in some cases). It prevents concurrent updates in a single jvm.

_This_ is the part I want to get rid of. The locking we do right now only ensures two things, neither of which we still need, and is preventing scaling the deployment. 
1. we only write one thing at a time (not particularly helpful, because two calls don't have to arrive concurrently to be trying to update the same thing, as just discussed)
2. that version increases uniformly (which we no longer need)

> So, i'm assuming we are good here. The concurrent update is whole different problem that needs to be handled separately right?

I was thinking we'd remove the single JVM locking and versioning, not just the history, but I'm ok doing one then the other as long as both get done in the near term, as I'd consider the two goals fairly equal in priority. The new feature to fail on blind overwrite (or whatever we'd like to call it) can and should be handled separately.

saxenakshitiz: Removed synchronized block

PR #112, opened 2022-05-12
refactor: replace lock with unique constraint

aaron-steinfeld:
## Description
Rather than using an in-memory lock, which prevents horizontal scaling, we're changing the label getOrCreate method here to retry. In conjunction with a separate change to add a unique constraint on the key, this should result in the lock being unnecessary. Concurrent calls may result in an error in the scenario where a label was checked for existence - returned false, then attempted to create and was created in the lag time. In this scenario, we attempt the method again under the assumption the initial read call will work the second time. If we can neither get nor create after two attempts, we give up. In no scenario do we ever end up with duplicates.

### Testing
TODO

### Checklist:

### DIFF
--- a/labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
+++ b/labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
@@ -1,9 +1,11 @@
 package org.hypertrace.label.config.service;
 
+import static com.github.rholder.retry.StopStrategies.stopAfterAttempt;
+import static com.github.rholder.retry.WaitStrategies.fixedWait;
 import static java.util.function.Function.identity;
 
-import com.google.common.util.concurrent.Striped;
-import com.google.protobuf.Duration;
+import com.github.rholder.retry.RetryerBuilder;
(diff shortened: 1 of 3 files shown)


labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -107,32 +103,23 @@ public void createLabel(
...
-          Label createdLabel = labelStore.upsertObject(requestContext, label).getData();
-          responseObserver.onNext(CreateLabelResponse.newBuilder().setLabel(createdLabel).build());
-          responseObserver.onCompleted();
-        } finally {
-          labelsLock.unlock();
-        }
-      } else {
-        responseObserver.onError(new StatusRuntimeException(Status.ABORTED));
+      LabelData labelData = request.getData();
+      if (isDuplicateKey(requestContext, labelData.getKey())) {
```
aaron-steinfeld, line 107: No longer strictly required, but I left these in place to get the better error message (vs getting the unique constraint violation)

aaron-steinfeld:
Snyk fixed in https://github.com/hypertrace/config-service/pull/113

PR #115, opened 2022-06-16
apis for inclusion rules and sampling config

SrikarMannepalli:


span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto
```diff
@@ -149,3 +150,46 @@ message ApiNamingRuleMetadata {
...
 }
+
+message SamplingConfig {
+  string id = 1;
+  Scope scope = 2;
+  RateLimitConfig rate_limit_config = 3;
+  SpanFilter filter = 4;
+  bool disabled = 5;
+}
+
```
SrikarMannepalli, line 161: The idea is to send the inclusion rules as disabled sampling config(with negated filters) with the highest priority and API level override scoping.

span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto
```diff
@@ -149,3 +150,46 @@ message ApiNamingRuleMetadata {
...
+    WindowedRateLimit fixed_window_limit = 1;
+  }
+}
+
+message WindowedRateLimit {
+  int64 quantity_allowed = 1;
+  google.protobuf.Duration window_duration = 2;
+}
+
+message IncludeSpanRule {
```
SrikarMannepalli, line 186: IncludeSpanRules currently look very similar to exclude span rules. Instead of `SpanFilter`, we could use something similar to the below as they would most likely be configured based on URL regex. But used `SpanFilter` as it is an umbrella over the below one and provides more flexibility.
```
message UrlRegexInfo {
   string regex = 1;
   string value = 2;
}
```

span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto
```diff
@@ -149,3 +150,46 @@ message ApiNamingRuleMetadata {
   google.protobuf.Timestamp creation_timestamp = 1;
   google.protobuf.Timestamp last_updated_timestamp = 2;
 }
+
+message SamplingConfig {
```
SrikarMannepalli, line 154: One issue that I'm contemplating is - If there are conflicting sampling configs at the same hierarchy(say we define 2 conflicting sampling configs for the same environment), how do we handle it ?🤔
Since, this override mechanism can be done by traceable admins alone, should we assume this would not be the case ?

span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto
SrikarMannepalli: The reasoning behind introducing disabled field is the following - If we are to send the inclusion rules modified as sampling config, we have to specify a way to say that the sampling rate(I mean per traces, not per time frame) must be 100%. Since we do not have this concept of rate(traces to sample/total) here(we specify traces to sample/time frame), added `disabled`.
`disabled` would be `false` for all the sampling configs except for the inclusion rules related ones, for which it would always be `true`.

SrikarMannepalli:
Discussed the approach offline. Moving these APIs from here.

skjindal93 (approved):

PR #122, opened 2022-07-06
refactor: use service factory pattern

aaron-steinfeld:
## Description
Switched to use config service factory. Improved health check.

### Testing
Verified local build, unit, integration tests.


surajpuvvada (dismissed):

tim-mwangi (dismissed):

aaron-steinfeld:
Requires https://github.com/hypertrace/service-framework/pull/44

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceModule.java
```diff
@@ -3,14 +3,14 @@
...
 import com.google.inject.Provides;
 import io.grpc.BindableService;
-import io.grpc.ManagedChannel;
+import io.grpc.Channel;
 import org.hypertrace.config.service.v1.ConfigServiceGrpc;
 import org.hypertrace.core.grpcutils.client.RequestContextClientCallCredsProviderFactory;
 
 public class SpanProcessingConfigServiceModule extends AbstractModule {
-  private final ManagedChannel channel;
+  private final Channel channel;
```
aaron-steinfeld, line 11: btw this is some tech debt I've been noticing that I've been correcting (since the framework does not expose the ManagedChannel). We don't need a managed channel for our purposes, any channel will do - and in fact we don't want one, because that would allow any consumer to manage them directly, affecting other consumers. The registry is the only place channels should be managed.

tim-mwangi: So this should be gotten rid off?

tim-mwangi (approved):

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceModule.java
aaron-steinfeld: Just switched from ManagedChannel to Channel as done here - as in all APIs, we want to accept the bare minimum that we need and all we need throughout application code (basically, anywhere except the registry that actually does the management) is a `Channel`.

PR #123, opened 2022-07-13
rename(span-processing-rules): documentation to spec id for API naming

skjindal93:


span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto
```diff
@@ -9,21 +9,29 @@ import "google/protobuf/duration.proto";
 import "google/protobuf/timestamp.proto";
 
 service SpanProcessingConfigService {
-  rpc CreateExcludeSpanRule(CreateExcludeSpanRuleRequest) returns (CreateExcludeSpanRuleResponse) {}
+  rpc CreateExcludeSpanRule (CreateExcludeSpanRuleRequest) returns (CreateExcludeSpanRuleResponse) {
```
skjindal93, line 12: That was my auto-indenter!

SrikarMannepalli (approved):

PR #124, opened 2022-07-13
Bulk create API for API naming rules

SrikarMannepalli:


span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto
```diff
@@ -100,6 +103,14 @@ message CreateApiNamingRuleResponse {
   ApiNamingRuleDetails rule_details = 1;
 }
 
+message CreateApiNamingRulesRequest {
+  repeated ApiNamingRuleInfo rules_info = 1;
+}
+
+message CreateApiNamingRulesResponse {
+  repeated ApiNamingRuleDetails rules_details = 1;
```
SrikarMannepalli, line 111: Couldn't think of a better field name. What do you suggest? `detailed_rules` maybe?🤔

skjindal93 (approved):

PR #127, opened 2022-07-18
add bulk update and delete apis for api naming rules

SrikarMannepalli:


span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -288,6 +295,45 @@ public void updateApiNamingRule(
...
+  @Override
+  public void updateApiNamingRules(
+      UpdateApiNamingRulesRequest request,
+      StreamObserver<UpdateApiNamingRulesResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      this.validator.validateOrThrow(requestContext, request);
+
+      List<UpdateApiNamingRule> updateApiNamingRules = request.getRulesList();
+      List<ApiNamingRule> existingRules =
```
skjindal93, line 307: Do we not have bulk get in config store?

skjindal93 (approved):

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
SrikarMannepalli: No Shubham.🥲

PR #131, opened 2022-08-10
de-duplication of api naming rules

SrikarMannepalli:


span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/validation/SpanProcessingConfigRequestValidator.java
```diff
@@ -145,12 +144,8 @@ private void validateConfig(ApiNamingRuleConfig ruleConfig) {
         validateRegex(segmentMatchingBasedConfig.getRegexesList());
         break;
       case API_SPEC_BASED_CONFIG:
+        // TODO: Add validations for specId list after migration of upstream services
         ApiSpecBasedConfig apiSpecBasedConfig = ruleConfig.getApiSpecBasedConfig();
-        if (Strings.isNullOrEmpty(apiSpecBasedConfig.getApiSpecId())) {
```
aman-bansal, line 149: shouldn't we make it either spec id and spec ids list should be there otherwise its an invalid request.

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/apinamingrules/DefaultApiNamingRulesManager.java
```diff
@@ -0,0 +1,269 @@
...
+  public void deleteApiNamingRule(
+      RequestContext requestContext, DeleteApiNamingRuleRequest request) {
+    // TODO: need to handle priorities
+    this.apiNamingRulesConfigStore
+        .deleteObject(requestContext, request.getId())
+        .orElseThrow(Status.NOT_FOUND::asRuntimeException);
+  }
+
+  @Override
+  public void deleteApiNamingRules(
```
aman-bansal, line 134: will it work? I remember the id gets map to context value in the config object. If you need to expose these api you should add this as attribute in the api naming rule itself

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/validation/SpanProcessingConfigRequestValidator.java
SrikarMannepalli: Right. Updated.

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/apinamingrules/DefaultApiNamingRulesManager.java
SrikarMannepalli: We are using `IdentifiedObjectStore` for api naming rules. The delete api of this object store just takes the id alone.

skjindal93 (approved):

PR #133, opened 2022-08-30
feat(system-level-exclusion-rules): system level exclude span rules

SrikarMannepalli:
## Description
This PR adds support for defining system level exclude span rules and overriding(enabling/disabling) them at a tenant level.

### Testing
Added new unit tests for all new code, and ran all the existing unit tests.

### Checklist:

### DIFF
--- a/span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
+++ b/span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
@@ -1,9 +1,18 @@
 package org.hypertrace.span.processing.config.service;
 
 import com.google.inject.Inject;
+import com.google.protobuf.util.JsonFormat;
+import com.typesafe.config.Config;
 import io.grpc.Status;
 import io.grpc.stub.StreamObserver;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.UUID;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import lombok.SneakyThrows;
 import lombok.extern.slf4j.Slf4j;
 import org.hypertrace.config.objectstore.ContextualConfigObject;
 import org.hypertrace.core.grpcutils.context.RequestContext;
@@ -43,21 +52,49 @@ import org.hypertrace.span.processing.config.service.validation.SpanProcessingCo
 @Slf4j
 class SpanProcessingConfigServiceImpl
     extends SpanProcessingConfigServiceGrpc.SpanProcessingConfigServiceImplBase {
+
+  private static final String SPAN_PROCESSING_CONFIG_SERVICE = "span.processing.config.service";
+  private static final String SYSTEM_LEVEL_EXCLUDE_SPAN_RULES = "system.level.exclude.span.rules";
+
   private final SpanProcessingConfigRequestValidator validator;
   private final ExcludeSpanRulesConfigStore excludeSpanRulesConfigStore;
   private final TimestampConverter timestampConverter;
   private final ApiNamingRulesManager apiNamingRulesManager;
+  private List<ExcludeSpanRule> systemLevelExcludeSpanRules;
+  private Map<String, ExcludeSpanRuleDetails> systemLevelExcludeSpanRulesToIdMap;
 
   @Inject
   SpanProcessingConfigServiceImpl(
       ExcludeSpanRulesConfigStore excludeSpanRulesConfigStore,
       SpanProcessingConfigRequestValidator requestValidator,
       TimestampConverter timestampConverter,
-      ApiNamingRulesManager apiNamingRulesManager) {
+      ApiNamingRulesManager apiNamingRulesManager,
+      Config config) {
     this.validator = requestValidator;
     this.excludeSpanRulesConfigStore = excludeSpanRulesConfigStore;
     this.timestampConverter = timestampConverter;
     this.apiNamingRulesManager = apiNamingRulesManager;
+
+    buildSystemLevelExcludeSpanRuleConfigs(config);
+  }
+
+  private void buildSystemLevelExcludeSpanRuleConfigs(Config config) {
+    List<? extends com.typesafe.config.ConfigObject> systemLevelExcludeSpanRuleObjectList = null;
+    if (config.hasPath(SPAN_PROCESSING_CONFIG_SERVICE)) {
+      Config spanProcessingConfig = config.getConfig(SPAN_PROCESSING_CONFIG_SERVICE);
+      if (spanProcessingConfig.hasPath(SYSTEM_LEVEL_EXCLUDE_SPAN_RULES)) {
+        systemLevelExcludeSpanRuleObjectList =
+            spanProcessingConfig.getObjectList(SYSTEM_LEVEL_EXCLUDE_SPAN_RULES);
+      }
+    }
+    if (systemLevelExcludeSpanRuleObjectList != null) {
+      systemLevelExcludeSpanRules =
+          buildSystemLevelExcludeSpanRules(systemLevelExcludeSpanRuleObjectList);
+      systemLevelExcludeSpanRulesToIdMap = buildSystemLevelExcludeSpanRulesToIdMap();
+    } else {
+      systemLevelExcludeSpanRules = Collections.emptyList();
+      systemLevelExcludeSpanRulesToIdMap = Collections.emptyMap();
+    }
   }
 
   @Override
@@ -67,10 +104,35 @@ class SpanProcessingConfigServiceImpl
     try {
       RequestContext requestContext = RequestContext.CURRENT.get();
       this.validator.validateOrThrow(requestContext, request);
+      List<ExcludeSpanRuleDetails> excludeSpanRules =
+          excludeSpanRulesConfigStore.getAllData(requestContext);
+
+      Map<String, ExcludeSpanRuleDetails> tenantLevelExcludeSpanRulesToIdMap =
+          buildTenantLevelExcludeSpanRulesToIdMap(excludeSpanRules);
+
+      // all the tenant level configs except for the overridden system level ones
+      List<ExcludeSpanRuleDetails> filteredTenantLevelExcludeSpanRules =
+          excludeSpanRules.stream()
+              .filter(
+                  excludeSpanRule ->
+                      !systemLevelExcludeSpanRulesToIdMap.containsKey(
+                          excludeSpanRule.getRule().getId()))
+              .collect(Collectors.toUnmodifiableList());
+
+      // all the system level configs, replaced by the overridden configs in case of overrides
+      List<ExcludeSpanRuleDetails> filteredSystemLevelExcludeSpanRules =
+          systemLevelExcludeSpanRules.stream()
+              .map(
+                  excludeSpanRule ->
+                      tenantLevelExcludeSpanRulesToIdMap.containsKey(excludeSpanRule.getId())
+                          ? tenantLevelExcludeSpanRulesToIdMap.get(excludeSpanRule.getId())
+                          : ExcludeSpanRuleDetails.newBuilder().setRule(excludeSpanRule).build())
+              .collect(Collectors.toUnmodifiableList());
 
       responseObserver.onNext(
           GetAllExcludeSpanRulesResponse.newBuilder()
-              .addAllRuleDetails(excludeSpanRulesConfigStore.getAllData(requestContext))
+              .addAllRuleDetails(filteredTenantLevelExcludeSpanRules)
+              .addAllRuleDetails(filteredSystemLevelExcludeSpanRules)
               .build());
       responseObserver.onCompleted();
     } catch (Exception e) {
@@ -116,9 +178,13 @@ class SpanProcessingConfigServiceImpl
       this.validator.validateOrThrow(requestContext, request);
 
       UpdateExcludeSpanRule updateExcludeSpanRule = request.getRule();
+
+      // check if the rule already exists. If not, check if it is a system level config. If yes, use
+      // that. Else, error out.
       ExcludeSpanRule existingRule =
           this.excludeSpanRulesConfigStore
               .getData(requestContext, updateExcludeSpanRule.getId())
+              .or(() -> getSystemLevelExcludeSpanRule(updateExcludeSpanRule.getId()))
               .orElseThrow(Status.NOT_FOUND::asException);
       ExcludeSpanRule updatedRule = buildUpdatedRule(existingRule, updateExcludeSpanRule);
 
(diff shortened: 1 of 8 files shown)


span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -43,21 +52,49 @@
 @Slf4j
 class SpanProcessingConfigServiceImpl
     extends SpanProcessingConfigServiceGrpc.SpanProcessingConfigServiceImplBase {
+
+  private static final String SPAN_PROCESSING_CONFIG_SERVICE = "span.processing.config.service";
+  private static final String SYSTEM_LEVEL_EXCLUDE_SPAN_RULES = "system.level.exclude.span.rules";
```
skjindal93, line 57: nit:
```suggestion
  private static final String DEFAULT_EXCLUDE_SPAN_RULES = "default.exclude.span.rules";
```

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -43,21 +52,49 @@
...
+
+  private void buildSystemLevelExcludeSpanRuleConfigs(Config config) {
+    List<? extends com.typesafe.config.ConfigObject> systemLevelExcludeSpanRuleObjectList = null;
+    if (config.hasPath(SPAN_PROCESSING_CONFIG_SERVICE)) {
+      Config spanProcessingConfig = config.getConfig(SPAN_PROCESSING_CONFIG_SERVICE);
+      if (spanProcessingConfig.hasPath(SYSTEM_LEVEL_EXCLUDE_SPAN_RULES)) {
+        systemLevelExcludeSpanRuleObjectList =
+            spanProcessingConfig.getObjectList(SYSTEM_LEVEL_EXCLUDE_SPAN_RULES);
+      }
+    }
```
skjindal93, line 89: Can we move the configs to a separate config class? Or, may be to a separate method. Would prefer a separate config class, if it isn't much refactor

We can also directly define the default exclusion rules config as `span.processing.config.service.system.exclude.span.rules`, to avoid multiple if block nesting

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -43,21 +52,49 @@
...
+
+  private static final String SPAN_PROCESSING_CONFIG_SERVICE = "span.processing.config.service";
+  private static final String SYSTEM_LEVEL_EXCLUDE_SPAN_RULES = "system.level.exclude.span.rules";
+
   private final SpanProcessingConfigRequestValidator validator;
   private final ExcludeSpanRulesConfigStore excludeSpanRulesConfigStore;
   private final TimestampConverter timestampConverter;
   private final ApiNamingRulesManager apiNamingRulesManager;
+  private List<ExcludeSpanRule> systemLevelExcludeSpanRules;
+  private Map<String, ExcludeSpanRuleDetails> systemLevelExcludeSpanRulesToIdMap;
```
skjindal93, line 64:
```suggestion
  private Map<String, ExcludeSpanRuleDetails> systemLevelExcludeSpanRuleIdToRuleMap;
```

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -43,21 +52,49 @@
...
     extends SpanProcessingConfigServiceGrpc.SpanProcessingConfigServiceImplBase {
+
+  private static final String SPAN_PROCESSING_CONFIG_SERVICE = "span.processing.config.service";
+  private static final String SYSTEM_LEVEL_EXCLUDE_SPAN_RULES = "system.level.exclude.span.rules";
+
   private final SpanProcessingConfigRequestValidator validator;
   private final ExcludeSpanRulesConfigStore excludeSpanRulesConfigStore;
   private final TimestampConverter timestampConverter;
   private final ApiNamingRulesManager apiNamingRulesManager;
+  private List<ExcludeSpanRule> systemLevelExcludeSpanRules;
```
skjindal93, line 63: Do we need this, if we already have the map?

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
skjindal93: Can we instead keep the map as `Map<String, ExcludeSpanRule>`? We can build the exclude span rule detail from exclude span rule at the end, while sending the response

I believe it would simplify the code. What say?

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -67,10 +104,35 @@ public void getAllExcludeSpanRules(
...
+                      tenantLevelExcludeSpanRulesToIdMap.containsKey(excludeSpanRule.getId())
+                          ? tenantLevelExcludeSpanRulesToIdMap.get(excludeSpanRule.getId())
+                          : ExcludeSpanRuleDetails.newBuilder().setRule(excludeSpanRule).build())
+              .collect(Collectors.toUnmodifiableList());
 
       responseObserver.onNext(
           GetAllExcludeSpanRulesResponse.newBuilder()
-              .addAllRuleDetails(excludeSpanRulesConfigStore.getAllData(requestContext))
+              .addAllRuleDetails(filteredTenantLevelExcludeSpanRules)
+              .addAllRuleDetails(filteredSystemLevelExcludeSpanRules)
```
skjindal93, line 135: This looks complex. I don't think there was a need to filter out overrides from persisted exclude span rules. We could have actually just sent out the entire persisted exclude span rules + non overriden system exclude rules

Something like this

```
Set<String> userExcludeSpanRuleIds = excludeSpanRules.stream()
    .map(excludeSpanRule -> excludeSpanRule.getRule().getId()).collect(
        Collectors.toUnmodifiableSet());

// all the system level configs, replaced by the overridden configs in case of overrides
List<ExcludeSpanRule> nonOverridenExcludeSpanRules =
    systemLevelExcludeSpanRules.stream()
        .filter(
            systemExcludeRule -> !userExcludeSpanRuleIds.contains(systemExcludeRule.getId()))
```

And, we can then convert the `nonOverridenExcludeSpanRules` to `List<ExcludeSpanRuleDetails>` while sending it in the response

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
skjindal93, line 57:
```suggestion
  private static final String SYSTEM_EXCLUDE_SPAN_RULES = "system.exclude.span.rules";
```

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -67,10 +104,35 @@ public void getAllExcludeSpanRules(
     try {
       RequestContext requestContext = RequestContext.CURRENT.get();
       this.validator.validateOrThrow(requestContext, request);
+      List<ExcludeSpanRuleDetails> excludeSpanRules =
+          excludeSpanRulesConfigStore.getAllData(requestContext);
+
+      Map<String, ExcludeSpanRuleDetails> tenantLevelExcludeSpanRulesToIdMap =
```
skjindal93, line 110: Instead of tenant level, can we call them user?
```suggestion
      Map<String, ExcludeSpanRuleDetails> userExcludeSpanRuleIdToRuleMap =
```

Though, I don't believe we even need to build the map

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -156,6 +222,47 @@ public void deleteExcludeSpanRule(
     }
   }
 
+  private List<ExcludeSpanRule> buildSystemLevelExcludeSpanRules(
```
skjindal93, line 225:
```suggestion
  private List<ExcludeSpanRule> buildSystemExcludeSpanRules(
```

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -156,6 +222,47 @@ public void deleteExcludeSpanRule(
...
+  @SneakyThrows
+  private ExcludeSpanRule buildExcludeSpanRuleFromConfig(
+      com.typesafe.config.ConfigObject configObject) {
+    String jsonString = configObject.render();
+    ExcludeSpanRule.Builder builder = ExcludeSpanRule.newBuilder();
+    JsonFormat.parser().merge(jsonString, builder);
+    return builder.build();
+  }
+
+  private Optional<ExcludeSpanRule> getSystemLevelExcludeSpanRule(String id) {
```
skjindal93, line 241:
```suggestion
  private Optional<ExcludeSpanRule> getSystemExcludeSpanRule(String id) {
```

skjindal93: Can we drop the term `level` across and call them system exclude span rule? Similary, other type of rules become user exclude span rule

skjindal93 (commented):
Can we add validation for not allowing updates for system level rules, apart from the disabled flag?

We should also add validation in delete. Deletion of system rules shouldn't be allowed. We can have this check within the delete API impl

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -116,9 +178,13 @@ public void updateExcludeSpanRule(
       this.validator.validateOrThrow(requestContext, request);
 
       UpdateExcludeSpanRule updateExcludeSpanRule = request.getRule();
+
+      // check if the rule already exists. If not, check if it is a system level config. If yes, use
+      // that. Else, error out.
       ExcludeSpanRule existingRule =
           this.excludeSpanRulesConfigStore
               .getData(requestContext, updateExcludeSpanRule.getId())
+              .or(() -> getSystemLevelExcludeSpanRule(updateExcludeSpanRule.getId()))
```
skjindal93, line 187: How are we taking care of the fact, if a customer disables the system exclude rule, we need to store is as disabled?

SrikarMannepalli: It is handled in building updated rule part on L189. The new rule would be built from the existing system exclude rule and the update exclude rule, and upserted.

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
SrikarMannepalli: I had taken this route so as to ensure that the system exclude rules(overridden or not) would end up in the bottom and the UI wouldn't have to worry about the ordering.

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
SrikarMannepalli: Right. Will update it.

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
SrikarMannepalli: Right. Updated to use the map in other usages as well.

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
SrikarMannepalli: I have added separate methods for reading configs and conversion for now. 
I am thinking of taking up a total refactor(config class, separate manager for exclude rules) separately. What do you suggest?

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -43,21 +57,42 @@
...
+    List<? extends com.typesafe.config.ConfigObject> systemExcludeSpanRuleObjectList = null;
+    if (config.hasPath(SYSTEM_EXCLUDE_SPAN_RULES)) {
+      systemExcludeSpanRuleObjectList = config.getObjectList(SYSTEM_EXCLUDE_SPAN_RULES);
+    }
+    if (systemExcludeSpanRuleObjectList != null) {
+      systemExcludeSpanRuleIdToRuleMap =
+          buildSystemExcludeSpanRulesToIdMap(systemExcludeSpanRuleObjectList);
+    } else {
+      systemExcludeSpanRuleIdToRuleMap = Collections.emptyMap();
+    }
```
skjindal93, line 95: nit: 
```suggestion
  private void buildSystemExcludeSpanRuleConfigs(Config config) {
    if (!config.hasPath(SYSTEM_EXCLUDE_SPAN_RULES)) {
      return;
    }

    List<? extends com.typesafe.config.ConfigObject> systemExcludeSpanRuleObjectList = config
        .getObjectList(SYSTEM_EXCLUDE_SPAN_RULES);
    systemExcludeSpanRuleIdToRuleMap =
        buildSystemExcludeSpanRulesToIdMap(systemExcludeSpanRuleObjectList);
  }
```

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/SpanProcessingConfigServiceImpl.java
```diff
@@ -156,6 +209,77 @@ public void deleteExcludeSpanRule(
...
+
+  private Optional<ExcludeSpanRule> getSystemExcludeSpanRule(String id) {
+    if (systemExcludeSpanRuleIdToRuleMap.containsKey(id)) {
+      return Optional.of(systemExcludeSpanRuleIdToRuleMap.get(id));
+    }
+    return Optional.empty();
+  }
+
+  private List<ExcludeSpanRuleDetails> reorderExcludeSpanRules(
+      List<ExcludeSpanRuleDetails> userExcludeSpanRules,
```
skjindal93, line 229: If we convert list into a map, then we can directly use `Maps.difference(userExcludeSpanRuleIdToRuleMap, systemExcludeSpanRuleIdToRuleMap)`. You can use

`entriesInCommon` for overriden rules
`entriesOnLeft` for only user rules
`entriesOnRight` for only system rules

SrikarMannepalli: But we would not be able to retain order among rules within the same groups right?🤔 Am I missing something?

skjindal93 (approved):

PR #134, opened 2022-09-01
feat: add config store abstractions for filter and scope resolution

singhalprerana:
Added abstractions over IdentifiedObjectStore with common logic to support filter and scope resolution

### DIFF
--- a/object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
+++ b/object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
@@ -1,0 +1,65 @@
+package org.hypertrace.config.objectstore;
+
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+import org.hypertrace.config.service.change.event.api.ConfigChangeEventGenerator;
+import org.hypertrace.config.service.v1.ConfigServiceGrpc;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+/**
+ * IdentifiedObjectStoreWithFilter is an abstraction over {@link IdentifiedObjectStore} to allow
+ * filtering on fetched configs
+ *
+ * @param <T> Config Object
+ * @param <F> Filter
+ */
+public abstract class IdentifiedObjectStoreWithFilter<T, F> extends IdentifiedObjectStore<T> {
+  protected IdentifiedObjectStoreWithFilter(
+      ConfigServiceGrpc.ConfigServiceBlockingStub configServiceBlockingStub,
+      String resourceNamespace,
+      String resourceName,
+      ConfigChangeEventGenerator configChangeEventGenerator) {
+    super(configServiceBlockingStub, resourceNamespace, resourceName, configChangeEventGenerator);
+  }
+
+  public List<T> getAllFilteredConfigs(RequestContext context, F filter) {
+    Stream<T> configObjectsStream = getAllObjects(context).stream().map(ConfigObject::getData);
+    if (!isFilterEmpty(filter)) {
+      configObjectsStream =
+          configObjectsStream
+              .map(configObject -> filterConfig(configObject, filter))
+              .filter(Optional::isPresent)
+              .map(Optional::get);
+    }
+    return configObjectsStream.collect(Collectors.toUnmodifiableList());
+  }
+
+  public Optional<T> getFilteredConfig(RequestContext context, String contextId, F filter) {
+    Optional<T> fetchedConfig = getData(context, contextId);
+    if (!isFilterEmpty(filter)) {
+      return fetchedConfig.flatMap(config -> filterConfig(config, filter));
+    }
+    return fetchedConfig;
+  }
+
+  /**
+   * Method to obtain filtered config A config object can consist of multiple configs, e.g. array,
+   * which can be partially filtered and returned. In cases where filter applies to entire config
+   * object, Optional.empty() is returned
+   *
+   * @param config
+   * @param filter
+   * @return
+   */
+  protected abstract Optional<T> filterConfig(T config, F filter);
+
+  /**
+   * Method to check if any filter confitions exist at all
+   *
+   * @param filter
+   * @return
+   */
+  protected abstract boolean isFilterEmpty(F filter);
+}
--- a/object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
+++ b/object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
@@ -1,0 +1,109 @@
+package org.hypertrace.config.objectstore;
+
+import com.google.protobuf.Descriptors;
+import com.google.protobuf.Message;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.hypertrace.config.service.change.event.api.ConfigChangeEventGenerator;
+import org.hypertrace.config.service.v1.ConfigServiceGrpc;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+/**
+ * ScopeResolutionIdentifiedObjectStoreWithFilter is an abstraction over {@link
+ * IdentifiedObjectStoreWithFilter} to allow scope resolution of fetched configs
+ *
+ * @param <T> Config Object
+ * @param <F> Filter
+ * @param <S> Scope
+ */
+public abstract class ScopeResolutionIdentifiedObjectStoreWithFilter<T extends Message, F, S>
+    extends IdentifiedObjectStoreWithFilter<T, F> {
+
+  protected ScopeResolutionIdentifiedObjectStoreWithFilter(
+      ConfigServiceGrpc.ConfigServiceBlockingStub configServiceBlockingStub,
+      String resourceNamespace,
+      String resourceName,
+      ConfigChangeEventGenerator configChangeEventGenerator) {
+    super(configServiceBlockingStub, resourceNamespace, resourceName, configChangeEventGenerator);
+  }
+
+  public List<T> getAllFilteredResolvedConfigs(RequestContext context, F filter, T defaultConfig) {
+    Map<S, T> configObjectsMap =
+        getAllFilteredConfigs(context, filter).stream()
+            .collect(Collectors.toMap(this::extractScope, Function.identity()));
+
+    return configObjectsMap.entrySet().stream()
+        .map(
+            entry -> resolveConfig(extractScope(entry.getValue()), defaultConfig, configObjectsMap))
+        .collect(Collectors.toUnmodifiableList());
+  }
+
+  public T getFilteredResolvedConfig(RequestContext context, F filter, S scope, T defaultConfig) {
+    List<T> filteredConfigObjects = getAllFilteredConfigs(context, filter);
+    Map<S, T> configObjectsMap =
+        filteredConfigObjects.stream()
+            .collect(Collectors.toMap(this::extractScope, Function.identity()));
+    return resolveConfig(scope, defaultConfig, configObjectsMap);
+  }
+
+  /**
+   * Method to extract scope object from given config object
+   *
+   * @param configObject
+   * @return Scope object
+   */
+  protected abstract S extractScope(T configObject);
+
+  /**
+   * Method to return scopes with which the given scope needs to be resolved in increasing order of
+   * priority For example, API scope would return a list {customer, env, service, api}
+   *
+   * @param scopeObject
+   * @return list of associated scopes in increasing order of priority
+   */
+  protected abstract List<S> getResolutionScopesWithIncreasingPriority(S scopeObject);
+
+  /**
+   * Method to merge 2 config objects Should be overridden in case the default merging logic doesn't
+   * apply, e.g. configs with arrays
+   *
+   * @param preferredConfig
+   * @param fallbackConfig
+   * @return mergedConfig
+   */
+  protected T mergeConfigs(T preferredConfig, T fallbackConfig) {
+    T.Builder builder = fallbackConfig.toBuilder();
+    mergeFromProto(builder, preferredConfig);
+    return (T) builder.build();
+  }
+
+  protected void mergeFromProto(T.Builder builder, T source) {
+    List<Descriptors.FieldDescriptor> fieldDescriptors = builder.getDescriptorForType().getFields();
+    for (Descriptors.FieldDescriptor fieldDescriptor : fieldDescriptors) {
+      if (fieldDescriptor.isRepeated()) {
+        builder.setField(fieldDescriptor, source.getField(fieldDescriptor));
+      } else if (source.hasField(fieldDescriptor)) {
+        if (fieldDescriptor.getJavaType() == Descriptors.FieldDescriptor.JavaType.MESSAGE) {
+          Message.Builder childBuilder = ((T) builder.getField(fieldDescriptor)).toBuilder();
+          mergeFromProto(childBuilder, (T) source.getField(fieldDescriptor));
+          builder.setField(fieldDescriptor, childBuilder.build());
+        } else {
+          builder.setField(fieldDescriptor, source.getField(fieldDescriptor));
+        }
+      }
+    }
+  }
+
+  private T resolveConfig(S scope, T defaultConfig, Map<S, T> configObjectsMap) {
+    List<S> scopesWithIncreasingPriority = getResolutionScopesWithIncreasingPriority(scope);
+    T resultConfig = defaultConfig;
+    for (S scopeObject : scopesWithIncreasingPriority) {
+      if (configObjectsMap.containsKey(scopeObject)) {
+        resultConfig = mergeConfigs(configObjectsMap.get(scopeObject), resultConfig);
+      }
+    }
+    return resultConfig;
+  }
+}


object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,65 @@
...
+    super(configServiceBlockingStub, resourceNamespace, resourceName, configChangeEventGenerator);
+  }
+
+  public List<T> getAllFilteredConfigs(RequestContext context, F filter) {
+    Stream<T> configObjectsStream = getAllObjects(context).stream().map(ConfigObject::getData);
+    if (!isFilterEmpty(filter)) {
+      configObjectsStream =
+          configObjectsStream
+              .map(configObject -> filterConfig(configObject, filter))
+              .filter(Optional::isPresent)
```
aaron-steinfeld, line 33: 1. filter + map here is equivalent to `flatMap(Optional::stream)`
2. we should be matching the non-filtered APIs (both in terms of name as well as signature), returning the wrapper objects. If we think it's useful to expose either, we could expose a second shortcut method.

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,65 @@
...
+   */
+  protected abstract Optional<T> filterConfig(T config, F filter);
+
+  /**
+   * Method to check if any filter confitions exist at all
+   *
+   * @param filter
+   * @return
+   */
+  protected abstract boolean isFilterEmpty(F filter);
```
aaron-steinfeld, line 64: Curious if we really need/want this. Usually, a caller should be using `hasFilter` style checks for presence detection, rather than handing off empty filters.

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,65 @@
...
+    Optional<T> fetchedConfig = getData(context, contextId);
+    if (!isFilterEmpty(filter)) {
+      return fetchedConfig.flatMap(config -> filterConfig(config, filter));
+    }
+    return fetchedConfig;
+  }
+
+  /**
+   * Method to obtain filtered config A config object can consist of multiple configs, e.g. array,
+   * which can be partially filtered and returned. In cases where filter applies to entire config
```
aaron-steinfeld, line 49: Can you point me to an example of a config that applies a filter internally like this?

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Map;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+import org.hypertrace.config.service.change.event.api.ConfigChangeEventGenerator;
+import org.hypertrace.config.service.v1.ConfigServiceGrpc;
+import org.hypertrace.core.grpcutils.context.RequestContext;
+
+/**
+ * ScopeResolutionIdentifiedObjectStoreWithFilter is an abstraction over {@link
+ * IdentifiedObjectStoreWithFilter} to allow scope resolution of fetched configs
```
aaron-steinfeld, line 15: I would have expected this to be an abstraction over DefaultObjectStore - typically if we're resolving scopes, we're not deal with IDs (think something like the booleans associated with detection policies). Or do we want to make that a requirement going forward? Even if we do, we'd still never likely fetch more than one at a time, so I don't think this inheritance relationship would make sense. will read on.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,109 @@
...
+
+  protected ScopeResolutionIdentifiedObjectStoreWithFilter(
+      ConfigServiceGrpc.ConfigServiceBlockingStub configServiceBlockingStub,
+      String resourceNamespace,
+      String resourceName,
+      ConfigChangeEventGenerator configChangeEventGenerator) {
+    super(configServiceBlockingStub, resourceNamespace, resourceName, configChangeEventGenerator);
+  }
+
+  public List<T> getAllFilteredResolvedConfigs(RequestContext context, F filter, T defaultConfig) {
```
aaron-steinfeld, line 32: Do we need to provide a default? That should likely be built into the store rather than provided on every call.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,109 @@
...
+   * @param fallbackConfig
+   * @return mergedConfig
+   */
+  protected T mergeConfigs(T preferredConfig, T fallbackConfig) {
+    T.Builder builder = fallbackConfig.toBuilder();
+    mergeFromProto(builder, preferredConfig);
+    return (T) builder.build();
+  }
+
+  protected void mergeFromProto(T.Builder builder, T source) {
```
aaron-steinfeld, line 82: Are we changing merge behavior such that we need to implement it custom? In other words, why can't we use the built in proto object merging, at least as the default.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,109 @@
...
+          builder.setField(fieldDescriptor, source.getField(fieldDescriptor));
+        }
+      }
+    }
+  }
+
+  private T resolveConfig(S scope, T defaultConfig, Map<S, T> configObjectsMap) {
+    List<S> scopesWithIncreasingPriority = getResolutionScopesWithIncreasingPriority(scope);
+    T resultConfig = defaultConfig;
+    for (S scopeObject : scopesWithIncreasingPriority) {
```
aaron-steinfeld, line 102: nit: this is just an imperative version of a stream reduce.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
aaron-steinfeld: What does it mean to get all resolved? Wouldn't we typically be either getting all unresolved, or getting a single resolved value?

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,109 @@
...
+        getAllFilteredConfigs(context, filter).stream()
+            .collect(Collectors.toMap(this::extractScope, Function.identity()));
+
+    return configObjectsMap.entrySet().stream()
+        .map(
+            entry -> resolveConfig(extractScope(entry.getValue()), defaultConfig, configObjectsMap))
+        .collect(Collectors.toUnmodifiableList());
+  }
+
+  public T getFilteredResolvedConfig(RequestContext context, F filter, S scope, T defaultConfig) {
```
aaron-steinfeld, line 43: Where does filter come in here? I think this comes back to using the filtered store as the base, but a call would typically have just a context and scope for the resolved version I'd expect. The scope acts as a filter.

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
singhalprerana: this can be a simple `hasCheck` too.. I didn't want the caller to decide between calling the filter api or regular one.. also, if filter isn't present, no point iterating thru the stream..

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
singhalprerana: `ai.traceable.anomaly.config.service.detector.anomalydetection.AnomalyDetectionConfigManagerImpl`

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: > Do we need to provide a default? That should likely be built into the store rather than provided on every call.
Yes agreed - will fix that

> What does it mean to get all resolved? Wouldn't we typically be either getting all unresolved, or getting a single resolved value?
Umm to reduce calls to config-service, we typically ask for all resolved configs - similar to fetch all rules.. example in same - `ai.traceable.anomaly.config.service.detector.anomalydetection.AnomalyDetectionConfigManager`

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: we don't use the built in proto object merging because that generally merges arrays while merging - which created issues some time back across multiple configs.. so we changed the behaviour using this custom logic to replace the arrays instead of merging..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: I thought I was missing something obvious but couldn't just place it.. thnx for pointing it out - will fix..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: umm no - as explained in above comment, 1 config object in a scope can have multiple configs and filter can be something other than scope too..

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
singhalprerana: 1 - yes will fix
2 - didn't understand.. do u mean the contextualWrapped objects should be returned instead? hmm was considering that but didn't really see many such usages.. but yeah, should keep APIs consistent - will fix

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
aaron-steinfeld: the hasCheck won't be possible since we've already had to call getFilter to get here and pass in the arg. If we don't want the caller to decide, we have a few options, none great:
1. remove any distinction on presence. In other words, an empty filter must be equivalent to an unset filter. This is what is being done here I believe.
2. pass in more context to this argument (e.g. passing in the request itself rather than - or instead of - just the filter allows the hasCheck). Downside here is that we typically have one store per resource but multiple methods (and thus multiple request shapes where that `hasCheck` might live.
3. skip the optimization (don't see a real reason to do this, at worst we could default it to false and allow opting in with an implementation if desired)

I guess after listing them out, callers that do have a distinction on presence can always opt out by putting their own check in front for example, so no real change requested here.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
aaron-steinfeld: oh gross - is this a copy paste from proto source? If so we should call that out and explain what is changed/why.

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
aaron-steinfeld: ahhh - this one is very different than pretty much every other config. So it gets read/written as one massive object? Was there a particular reason it was done this way, or is it just tech debt?

Will read on, but if this is the only config that buries multiple independent configs in one object, we can/should consider giving it a custom store so as not to affect the API for the straightforward cases.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
aaron-steinfeld: Ah, this is the change that got added for the use case where one of the streaming apps had been fetching/resolving each config per API and instead changed to fetch for all APIs for a given customer right? Makes sense as a use case, but we'd likely want both - getting all resolved (for the config consumer/applier) as well as getting all unresolved (for the config reader/writer ).

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
singhalprerana: fixed

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStoreWithFilter.java
singhalprerana: hmm.. removing

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: but without contexts how are the configs stored? they still have contexts - scopes define the ids in most cases.. i default-store, there is a single config per resource, I don't think that would work multiple scopes right?

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: getting all unresolved is anyways possible using methods from the super-class

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: not exactly copy-paste but a different implementation to replace the list instead of appending - added comment explaining that

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: fixed

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
aaron-steinfeld: I think the distinction is that they don't have explicit IDs, but the scope would end up being used as the context. I'm not positive that's needed/wanted as you indicated, so we can skip for now and come back if we find some use case.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
aaron-steinfeld: True, but one thing we have to be careful of is that we're correctly maintaining is-a relationships. For example, if we're resolving a config to produce an object that is a composite of multiple stored objects, it doesn't have its own ID, create time etc. It's a different object shape then the unresolved ones and doesn't support direct deleting or updating. This concern applies at the API level too, so not unique here.

One way would be to add a second type parameter for the resolved vs the persisted objects. Another would be to use a different store that only supports resolving (i.e. doesn't extend existing ones but potentially uses them internally) and the service can use a separate resolved and unresolved store.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,138 @@
...
+      T defaultConfig) {
+    super(configServiceBlockingStub, resourceNamespace, resourceName, configChangeEventGenerator);
+    this.defaultConfig = defaultConfig;
+  }
+
+  public T getDefaultConfig() {
+    return defaultConfig;
+  }
+
+  public List<ContextualConfigObject<T>> getAllFilteredResolvedObjects(
```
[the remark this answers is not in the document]
singhalprerana: Yes intend is to get all configs - # of configs returned = # of configs stored.. Just to be clear, if there are 4 configs stored - customer C1, service S1 and apis A1 and A2, we resolve all of them and return - C1R=C1+default, S1R=S1+C1+default, A1R=A1+S1+C1+default, A2R=A2+S1+C1+default (+ meaning resolved with)

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,138 @@
...
+  public List<T> getAllFilteredResolvedConfigData(RequestContext context, F filter) {
+    return getAllFilteredResolvedObjects(context, filter).stream()
+        .map(ConfigObject::getData)
+        .collect(Collectors.toUnmodifiableList());
+  }
+
+  public Optional<ContextualConfigObject<T>> getFilteredResolvedObject(
+      RequestContext context, F filter, S scope) {
+    Map<S, ContextualConfigObject<T>> configObjectsMap = getConfigObjectsMap(context, filter);
+    return Optional.ofNullable(configObjectsMap.get(scope))
```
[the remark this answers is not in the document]
singhalprerana: if there is no config stored for A3, configStore has no idea of existence of api A3.. It's only going to return the resolved version of the configs stored..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,138 @@
...
+        .map(
+            configObject ->
+                ContextualConfigObjectImpl.updateData(
+                    configObject, resolveConfig(scope, configObjectsMap)));
+  }
+
+  public Optional<T> getFilteredResolvedData(RequestContext context, F filter, S scope) {
```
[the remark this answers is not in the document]
singhalprerana: There are cases for both filter and scope - e.g. training configs, detection configs, etc..
we can have another oevrloaded method with no filter if u see that use-case..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: umm not exactly.. the library code is basically `mergeFrom` for individual protos which is very different for each proto msg.. this is not exactly a copy paste but a recursive workaround (which I think @kamaleshnneerasa had got from some stackoverflow solution)

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: yeah for now, I didn't see any req as such so skipping..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
```diff
@@ -0,0 +1,138 @@
...
+      ConfigServiceGrpc.ConfigServiceBlockingStub configServiceBlockingStub,
+      String resourceNamespace,
+      String resourceName,
+      ConfigChangeEventGenerator configChangeEventGenerator,
+      T defaultConfig) {
+    super(configServiceBlockingStub, resourceNamespace, resourceName, configChangeEventGenerator);
+    this.defaultConfig = defaultConfig;
+  }
+
+  public T getDefaultConfig() {
```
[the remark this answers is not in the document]
singhalprerana: had thought of keeping as optional but was facing an issue with merging logic in reduce method.. if the caller has no default, they can pass the default instance of the config object..
but removing the getter nevertheless..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
aaron-steinfeld: > umm services which require resolved objects also need unresolved ones 

Not sure I follow this? Do you mean the typed config services? If so, agreed. My point was just that the api to the resolved and unresolved ones is a bit different. In this PR, we add new request arguments to the API, but they share the same response type (driven by the type parameter `T`). For some APIs, the resolved and unresolved objects may both be T, for others, they may be different objects (for example, one without an "id" field). If we take either of the suggestions, they can account for this by not requiring `T` be the same in both cases.

singhalprerana: if I understand u correctly, what u mean is resolved config type might be different than `T` - I haven't seen such use-cases yet.. in fact the `id` can still be the same as the unresolved config - that's how it would be referenced anyways, especially in case of `getAll`.. having said that, if there is a need of such use-case later, can always add a separate store with different types for resolved and unresolved..
or alternatively can make this more generic as - `ScopeResolutionIdentifiedObjectStoreScopeResolutionIdentifiedObjectStore<T extends Message, U extends Message, F, S>` - though IMO not required for now..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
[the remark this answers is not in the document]
singhalprerana: I don't think we would be breaking APIs with this - at most, might need a constructor without the same.. but having said that, I think default config would be reqd - merging would always need one, at least the default instance.. but if u feel strongly, can have it as optional right away too (though that might be overkill for now)

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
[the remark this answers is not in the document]
singhalprerana: oh it will return that in this impl too - I misunderstood ur initial question.. what I explained was the impl for `getAll`.. 
if a call is made for getResolved for A3 and its config is not present, this impl should surely return config for S1 - forgot to use the chain call.. will fix..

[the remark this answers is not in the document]
singhalprerana: update - just checked.. the impl will return S1 - the `resolveConfig` method calls the chain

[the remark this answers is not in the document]
aaron-steinfeld: Maybe I'm looking at old code, but we don't call `resolveConfig` unless the requested scope is in the map. If it's a scope no persisted config has, then we'll always return empty

```java
Optional.ofNullable(configObjectsMap.get(scope))
```

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
aaron-steinfeld: > if I understand u correctly, what u mean is resolved config type might be different than T

Correct

> I haven't seen such use-cases yet.. in fact the id can still be the same as the unresolved config - that's how it would be referenced anyways, especially in case of getAll

I would call this a problem. We shouldn't be referencing a resolved config, only evaluating it. If we give it the ID of only one of its component parts, not only is that inaccurate, but that makes it really easy to try to for example upsert the resolved config and break the resolution. Resolved configs should ideally be different messages, even if they're not today, so we'd want to guide new APIs in that direction.

> if there is a need of such use-case later, can always add a separate store with different types for resolved and unresolved..
or alternatively can make this more generic as - ScopeResolutionIdentifiedObjectStoreScopeResolutionIdentifiedObjectStore<T extends Message, U extends Message, F, S> - though IMO not required for now..

So the latter was my second suggestion (and can easily assign T and U to the same thing if needed), the former is similar to my first suggestion, but If we're resolving from a separate store, then that's not a store for writing so I was instead suggesting decoupling the hierarchy rather then extending the existing stores.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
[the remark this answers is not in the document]
singhalprerana: yes old code - had missed it in 1 method, now fixed..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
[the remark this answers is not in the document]
singhalprerana: yeah shouldn't break the API as it's not in API anymore - only as part of constructor..

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
singhalprerana: I still don't see merit in decoupling the hierarchy - it should still extend the existing stores because the written ones and the unresolved ones (which is the input for resolving) would be the same - adding a separate type (U) for resolved one on top of it might make sense..

aaron-steinfeld: It's two separate sets of APIs. The existing stores provide basic CRUD without any real logic, while the resolution stuff is logic built on top of that basic CRUD to produce new objects (which aren't themselves writable). 

So extending the existing stores and allowing a separate type parameter allows us to put both sets of APIs in the same interface without requiring that they return the same type. This is a bit more convenient for callers since they don't have to manage a separate object (a CRUD store and a resolver - because you'd never use resolution without CRUD as well) but is also pushing the boundaries of/breaking single responsibility principle.

object-store/src/main/java/org/hypertrace/config/objectstore/ScopeResolutionIdentifiedObjectStoreWithFilter.java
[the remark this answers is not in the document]
aaron-steinfeld: Maybe a missed push? Looks like the latest still has this:
https://github.com/hypertrace/config-service/pull/134/files#diff-beb6243dae76a59024110378831901720eda9039152e42b1bde6f1ff615d2957R51-R53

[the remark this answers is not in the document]
singhalprerana: yeah - fixed..

object-store/src/main/java/org/hypertrace/config/objectstore/ConfigObjectScopeResolver.java
```diff
@@ -0,0 +1,131 @@
...
+   * Method to extract scope object from given config object
+   *
+   * @param configData
+   * @return Scope object
+   */
+  protected abstract S extractScope(T configData);
+
+  /**
+   * Method to return scopes with which the given scope needs to be resolved in increasing order of
+   * priority For example, API scope would return a list {customer, env, service, api}
```
aaron-steinfeld, line 52: nit
```suggestion
   * priority For example, API scope could return a list {tenant, service, api}
```

Keeping this relevant to HT, also theres no requirement that a service has to live above an api, for example, so just loosening the wording slightly.

object-store/src/main/java/org/hypertrace/config/objectstore/ConfigObjectScopeResolver.java
```diff
@@ -0,0 +1,131 @@
...
+        .collect(Collectors.toMap(this::extractScope, Function.identity()));
+  }
+
+  private Optional<U> resolveConfig(S scope, Map<S, T> configDataMap) {
+    Optional<T> mergedConfig =
+        getResolutionScopesWithIncreasingPriority(scope).stream()
+            .flatMap(scopeObject -> Optional.ofNullable(configDataMap.get(scope)).stream())
+            .reduce(
+                (fallbackConfig, preferredConfig) ->
+                    (T) mergeConfigs(fallbackConfig, preferredConfig));
```
aaron-steinfeld, line 87: both inputs are T right? we should fix the typing on the method such that both args and the response are same type without a cast. Something like
```java
 private <M extends Message> M mergeConfigs(M fallbackConfig, M preferredConfig) {
```

object-store/src/main/java/org/hypertrace/config/objectstore/ConfigObjectScopeResolver.java
```diff
@@ -0,0 +1,131 @@
...
+    return mergedConfig
```
aaron-steinfeld, line 88: It feels a bit strange to merge the stored configs via algo, then ask the implementer to implement how to merge the merged + default configs. What do you think of removing `protected abstract U resolveConfig(U fallbackConfig, T preferredConfig);` and converting to U before merging so we can do a single reduce via generic merge? something like
```
Stream.concat(
  defaultConfig.stream(), 
  getResolutionScopesWithIncreasingPriority(scope).stream().map(this::convertConfig)
).reduce(...)`

Comment (singhalprerana) [reply object-store/src/main/java/org/hypertrace/config/objectstore/ConfigObjectScopeResolver.java:88]: no wouldn't work in many cases because in many cases, we store configs with optional values in config store while the resolved value returned would need all values present.. for example,

```
msg T {
 optional bool disabled = 1;
 optional bool blocking = 2;
 optional string category = 3;
}
msg U {
 bool disabled = 1;
 bool blocking = 2;
 string category = 3;
}
```

Now let's say we have configs stored as `T-api1={disabled=true}, T-service1={blocking=true}` and default config as `U-default={disabled=false, blocking=false, category=latent}`.. now we want resolved config as `U-api1-resolved={disabled=true, blocking=true, category=latent}`.. but if we convert them individually to U, we would get `U-api1={disabled=true, blocking=false, category=latent}, T-service1={disabled=false, blocking=true, category=latent}` and merging them would result in `U-api1-resolved={disabled=true, blocking=false, category=latent}` because now all values are set for api1, not taking the fallback values from service1..

1 idea I was thinking about was not having `protected abstract U resolveConfig(U fallbackConfig, T preferredConfig);` - only keeping the `convertConfig()` method and caller can choose to use/not use default config to convert.. but that might be misleading, in case they want to convert any `T` config to `U` config independently too (sometimes very useful).. so kept these 2 different..

aaron-steinfeld: I think I'm following - definitely on the first part at least. The part I'm struggling with is - if I have to implement the accumulator(U + T -> U) function, what's the point of the out of the box merge? The accumulator could just as easily do the work. 

So if we want to use generic merging logic: 

We need to get each value in the merge chain to be the same type. Your explanation on why that shouldn't be the returned type U makes sense, so I would guess this would look like changing the default to be of type T, merging generically then using convert (T -> U) with the result. Throw if no values in merge chain (which means no default given).

If we don't and want to use an accumulator instead:

This is similar to what's in current PR. We need a starting point, so we require exactly one of [default (U) OR converter (T->U)] AND accumulator (U + T -> U). This case only throws if no default or values (no T to convert).  Basically, we don't need both ways of seeding the reduce, nor do we need the generic merge logic if we go this route, at least that's my understanding.

singhalprerana: This is complicating matters and might not solve the reqs.. changed the default config type to `T` - that might make more sense and simplifies the logic too..

aaron-steinfeld (approved):

PR #135, opened 2022-09-05
Added id field in webhookChannelConfig proto

sarthak77:


avinashkolluru (approved):

notification-channel-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -25,6 +25,7 @@ message WebhookChannelConfig {
   string url = 1;
   WebhookFormat format = 2;
   repeated WebhookHeader headers = 3;
+  string id = 4;
```
sarthak77, line 28: @aaron-steinfeld @avinashkolluru do we need to mark this as deprecated now?

aaron-steinfeld: This one's internal, if it's not used we can remove it. If it is, we should remove there first. The other thing to be careful of is make sure that the converter is ok dropping unknown properties, since there could be persisted IDs. A quick test should suffice on this front.

sarthak77: >if it's not used we can remove it

This was added for that secret header use case only. Since we have stopped using it there, I think we can even remove it. 

>make sure that the converter is ok dropping unknown properties, since there could be persisted IDs. A quick test should suffice on this front.

Didn't get this.

aaron-steinfeld: if a value is persisted in a database like `{url: "...", format: SOME_FORMAT, id: "xyz"}`, we want to make sure when retrieving that from the database it silently ignores the ID rather than failing to deserialize due to an unknown proeprty.

sarthak77: Currently also it is not being used anywhere. The PR in config service consuming this was reverted earlier. Other consumption in gql was overridden with that make shift id.

aaron-steinfeld: If it was used once (thus there's a value in the database), it doesn't matter that it was later unhooked. Please add a test to https://github.com/hypertrace/config-service/pull/141- it should be straightforward.

sarthak77: How do we test this? In the test how do we add a value with old proto and check retrieval using new proto?

aaron-steinfeld: Generate/write a notification channel object in JSON that has the ID field in it (can manually add this field or use the pre-revert code to generate it), then deserialize it into a `Value` and then test the method in this module - https://github.com/hypertrace/config-service/blob/795fb64a446ad9fc1e25cedd3c21f5f0707bd061/notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelStore.java#L35 to see that it correctly ignores that id field on deserialization and produces the object you expect.

sarthak77: Added.

PR #137, opened 2022-09-07
Populated id in create and update channel request

sarthak77:


notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceImpl.java
```diff
@@ -149,4 +151,19 @@ public void getNotificationChannel(
       responseObserver.onError(e);
     }
   }
+
+  private NotificationChannelMutableData getIdPopulatedData(NotificationChannelMutableData data) {
```
avinashkolluru, line 155: Shouldn't the same be done in update too. That way, overtime if anyone updates existing configs, they also have the ID.

sarthak77: This is being called from both update and create.

avinashkolluru (approved):

PR #138, opened 2022-09-12
Added logic to fetch secret headers

sarthak77:


notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceImpl.java
```diff
@@ -30,12 +29,14 @@
 
   private final NotificationChannelStore notificationChannelStore;
   private final NotificationChannelConfigServiceRequestValidator validator;
+  private final Utils utils;
```
aaron-steinfeld, line 32: Name specifically. Utils classes should be targeted.

notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceImpl.java
```diff
@@ -30,12 +29,14 @@
...
 
   public NotificationChannelConfigServiceImpl(
       Channel channel, ConfigChangeEventGenerator configChangeEventGenerator) {
     this.notificationChannelStore =
         new NotificationChannelStore(channel, configChangeEventGenerator);
     this.validator = new NotificationChannelConfigServiceRequestValidator();
+    this.utils = new Utils(this.notificationChannelStore);
```
aaron-steinfeld, line 39: Are we not using guice? (If no, don't need to add it here, if yes, please use guice to instantiate instead).

notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceImpl.java
```diff
@@ -78,7 +79,10 @@ public void updateNotificationChannel(
                           NotificationChannel.newBuilder()
                               .setId(request.getId())
                               .setNotificationChannelMutableData(
-                                  getIdPopulatedData(request.getNotificationChannelMutableData()))
+                                  utils.getHeaderPopulatedData(
```
aaron-steinfeld, line 82: Not sure what this means - please rename. Also if this is the only invocation, likely makes sense to move the call to the other util internal.

notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/Utils.java
```diff
@@ -0,0 +1,107 @@
...
+import org.hypertrace.notification.config.service.v1.NotificationChannel;
+import org.hypertrace.notification.config.service.v1.NotificationChannelMutableData;
+import org.hypertrace.notification.config.service.v1.WebhookChannelConfig;
+import org.hypertrace.notification.config.service.v1.WebhookHeader;
+
+public class Utils {
+
+  private final NotificationChannelStore notificationChannelStore;
+
+  public Utils(NotificationChannelStore notificationChannelStore) {
```
aaron-steinfeld, line 19: The fact that this takes in a store class is a major red flag. Utils are typically pure functions - data in, data out. No state or state changes.

Perhaps this indicates a better place for this logic would be in the store itself? For things like defaulting newly added fields, I would say that's for sure the case. Will read on to see what else this is doing.

sarthak77:
Closing this PR based on discussions here : https://github.com/Traceableai/graphql-service/pull/1100#discussion_r965439151
CC @aaron-steinfeld @avinashkolluru

PR #144, opened 2022-10-19
fixed vulnerabilitiy

bhubam:
## Description
Upgraded the protobuf-java-util version to remove the vulnerabilites

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,5 +1,5 @@
 [versions]
-protoc = "3.21.4"
+protoc = "3.21.7"
 grpc = "1.48.0"
 gson = "2.9.0"
 hypertrace-grpcutils = "0.8.2"
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -2,7 +2,7 @@
 protoc = "3.21.7"
 grpc = "1.50.0"
 gson = "2.9.0"
-hypertrace-grpcutils = "0.8.2"
+hypertrace-grpcutils = "0.9.0"
 hypertrace-framework = "0.1.43"
 lombok = "1.18.22"
 jackson = "2.13.2.2"
@@ -65,4 +65,4 @@ mockito-junit = { module = "org.mockito:mockito-junit-jupiter", version.ref = "m
 mockito-inline = { module = "org.mockito:mockito-inline", version.ref = "mockito" }
 
 [bundles]
-grpc-api = ["grpc-stub", "grpc-protobuf", "javax-annotation"]
+grpc-api = ["grpc-stub", "grpc-protobuf", "javax-annotation", "protobuf-java"]
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,6 +1,6 @@
 [versions]
 protoc = "3.21.7"
-grpc = "1.48.0"
+grpc = "1.50.0"
 gson = "2.9.0"
 hypertrace-grpcutils = "0.8.2"
 hypertrace-framework = "0.1.43"


gradle/libs.versions.toml
```diff
@@ -1,6 +1,6 @@
 [versions]
-protoc = "3.21.4"
-grpc = "1.48.0"
+protoc = "3.21.7"
+grpc = "1.50.0"
```
avinashkolluru, line 3: So 3.21.7 version of protobuf-java-util is not compatible with 1.48 of grpc?
If thats the case @aaron-steinfeld should we be upgrading the grpc version directly here - https://github.com/hypertrace/java-grpc-utils/blob/main/grpc-context-utils/build.gradle.kts?

bhubam: Yeah 3.21.7  protobuf-java-util version is not compatible with 1.48.0 version of grpc, we have to use version >=1.48.2

aaron-steinfeld: Can you elaborate on why we think it's incompatible? It would be **very** surprising if proto introduced a breaking change.

aaron-steinfeld: (We can update the grpc version too - it actually shouldn't even be declared in repos any more, just inherited from the common lib - but it shouldn't be required so want to understand that first)

bhubam: Upgrading the protobuf-java-util version to 3.21.7 was failing the build. So I updated the grpc version, was reieving the following error: `[2022-10-19 06:10:08] [autobuild] /home/runner/work/config-service/config-service/alerting-config-service-api/build/generated/source/proto/main/java/org/hypertrace/alerting/config/service/v1/EventCondition.java:448: error: cannot find symbol
  [2022-10-19 06:10:08] [autobuild]               if (!super.parseUnknownField(input, extensionRegistry, tag)) {
  [2022-10-19 06:10:08] [autobuild]                         ^
  [2022-10-19 06:10:08] [autobuild]   symbol: method parseUnknownField(CodedInputStream,ExtensionRegistryLite,int)
  [2022-10-19 06:10:08] [autobuild] /home/runner/work/config-service/config-service/alerting-config-service-api/build/generated/source/proto/main/java/org/hypertrace/alerting/config/service/v1/EventConditionMutableData.java:447: error: cannot find symbol
  [2022-10-19 06:10:08] [autobuild]               if (!super.parseUnknownField(input, extensionRegistry, tag)) {
  [2022-10-19 06:10:08] [autobuild]                         ^
  [2022-10-19 06:10:08] [autobuild]   symbol: method parseUnknownField(CodedInputStream,ExtensionRegistryLite,int)
  [2022-10-19 06:10:08] [autobuild] /home/runner/work/config-service/config-service/alerting-config-service-api/build/generated/source/proto/main/java/org/hypertrace/alerting/config/service/v1/MetricAnomalyEventCondition.java:566: error: cannot find symbol`

aaron-steinfeld: Got it. So the problem is slightly different.

The error you're seeing is because you're updating proto in some places, but not in others. The generated files are now being generated with 3.21.7, but the api packages are resolving 3.21.1 transitively, via grpc:
image

Updating grpc brings in the newer transitive will solve the issue. But alternatively, so will declaring the version of proto where we're relying on it being the same. IMO we should do both (updating grpc because it's good to stay up to date, and that'll make other repos "just work"; declaring proto because we actually want to ensure the runtime is in sync with the compiled version rather than relying on using a matching grpc-proto version pair).

So tl;dr would suggest below diff after I cut 0.8.3 grpc utils:
```diff
diff --git a/gradle/libs.versions.toml b/gradle/libs.versions.toml
index fd0c2d7..8285c59 100644
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -1,8 +1,8 @@
 [versions]
-protoc = "3.21.4"
-grpc = "1.48.0"
+protoc = "3.21.7"
+grpc = "1.50.0"
 gson = "2.9.0"
-hypertrace-grpcutils = "0.8.2"
+hypertrace-grpcutils = "0.8.3"
 hypertrace-framework = "0.1.43"
 lombok = "1.18.22"
 jackson = "2.13.2.2"
@@ -65,4 +65,4 @@ mockito-junit = { module = "org.mockito:mockito-junit-jupiter", version.ref = "m
 mockito-inline = { module = "org.mockito:mockito-inline", version.ref = "mockito" }
 
 [bundles]
-grpc-api = ["grpc-stub", "grpc-protobuf", "javax-annotation"]
+grpc-api = ["grpc-stub", "grpc-protobuf", "javax-annotation", "protobuf-java"]

```

aaron-steinfeld: Merged this as `0.9.0` (as opposed to `0.8.3` - there were more significant changes, but all should be backwards compatible)

bhubam: Made the suggested changes, Please look

aaron-steinfeld (approved):

PR #145, opened 2022-12-09
chore: upgrade buf

aaron-steinfeld:
## Description
Upgrade to release buf version.


span-processing-config-service-api/src/main/proto/org/hypertrace/span/processing/config/service/v1/span_processing_config_service.proto
```diff
@@ -5,7 +5,6 @@ package org.hypertrace.span.processing.config.service.v1;
 option java_multiple_files = true;
 
 import "org/hypertrace/span/processing/config/service/v1/span_filter.proto";
-import "google/protobuf/duration.proto";
```
aaron-steinfeld, line 8: only actual change, unused import caught by improved linter.

.github/workflows/pr-build.yml
```diff
@@ -58,10 +58,18 @@ jobs:
...
+      - name: Setup buf
+        uses: bufbuild/buf-setup-action@v1.9.0
+        with:
+          github_token: ${{ github.token }}
+
+      - name: Lint protos
+        uses: bufbuild/buf-lint-action@v1
+
+      - name: Check for breaking changes
+        uses: bufbuild/buf-breaking-action@v1
```
aaron-steinfeld, line 70: These pre-packaged actions will provide better messaging/annotations directly on the PR.

aaron-steinfeld:
Expected validate protos to fail here for two different reasons:
- The way CI is set up in this repo, it uses the new code with the CI config from main (for security purposes). So CI is trying to give the new buf config files to the old buf tool.
- Even if not (I have a commit where I switched this up), the old version of buf generates a different output for comparison.

So tl;dr ignore the build failure, we'll override.

buf.work.yaml
```diff
@@ -0,0 +1,10 @@
+version: v1
```
aaron-steinfeld, line 1: These configs generated with `buf beta migrate-v1beta1` as described in https://docs.buf.build/configuration/v1beta1-migration-guide

tim-mwangi (approved):

PR #148, opened 2023-01-03
chore: use document store bulk upsert

SrikarMannepalli:
## Description
This PR refactors upsertAllConfigs API impl to use bulk upsert document store method

### Testing
Added unit tests and integration test to successfully verify the changes

### Checklist:

### DIFF
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
@@ -77,29 +77,11 @@ public class DocumentConfigStore implements ConfigStore {
     Optional<ConfigDocument> previousConfigDoc = getLatestVersionConfigDoc(configResourceContext);
     Optional<ContextSpecificConfig> optionalPreviousConfig =
         previousConfigDoc.flatMap(this::convertToContextSpecificConfig);
-    long updateTimestamp = clock.millis();
-    long creationTimestamp =
-        previousConfigDoc
-            .filter(configDocument -> !ConfigServiceUtils.isNull(configDocument.getConfig()))
-            .map(ConfigDocument::getCreationTimestamp)
-            .orElse(updateTimestamp);
-    long newVersion =
-        previousConfigDoc
-            .map(ConfigDocument::getConfigVersion)
-            .map(previousVersion -> previousVersion + 1)
-            .orElse(1L);
+
     Key latestDocKey = new ConfigDocumentKey(configResourceContext);
     ConfigDocument latestConfigDocument =
-        new ConfigDocument(
-            configResourceContext.getConfigResource().getResourceName(),
-            configResourceContext.getConfigResource().getResourceNamespace(),
-            configResourceContext.getConfigResource().getTenantId(),
-            configResourceContext.getContext(),
-            newVersion,
-            userId,
-            latestConfig,
-            creationTimestamp,
-            updateTimestamp);
+        buildConfigDocument(configResourceContext, latestConfig, userId, previousConfigDoc);
+
     collection.upsert(latestDocKey, latestConfigDocument);
     return optionalPreviousConfig
         .map(previousConfig -> this.buildUpsertResult(latestConfigDocument, previousConfig))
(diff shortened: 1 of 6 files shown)


config-service/build.gradle.kts
```diff
@@ -55,6 +55,7 @@ tasks.integrationTest {
 }
 
 dependencies {
+  implementation(libs.hypertrace.grpcutils.context)
```
aaron-steinfeld, line 58: Considering no source was changed in this module, shouldn't be any dep additions. Looks like maybe this was meant for the integration test?

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
```diff
@@ -162,68 +163,63 @@ void writeAllConfigs() throws IOException {
...
+    when(collection.bulkUpsert(any())).thenReturn(true);
     when(this.mockClock.millis()).thenReturn(updateTime);
     List<UpsertedConfig> upsertedConfigs =
         // Swap configs between contexts as an update
         configStore.writeAllConfigs(
             ImmutableMap.of(resourceContext1, config2, resourceContext2, config1), USER_ID);
 
-    assertEquals(
-        List.of(
+    assertEquals(2, upsertedConfigs.size());
```
aaron-steinfeld, line 186: Do we no longer have a guarantee of order? I wouldn't expect the api contract to change for an impl detail.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -76,6 +77,56 @@ public UpsertedConfig writeConfig(
...
+    return optionalPreviousConfig
+        .map(previousConfig -> this.buildUpsertResult(latestConfigDocument, previousConfig))
+        .orElseGet(() -> this.buildUpsertResult(latestConfigDocument));
+  }
+
+  private List<UpsertedConfig> writeConfigs(
+      Map<ConfigResourceContext, Value> resourceContextValueMap, String userId) throws IOException {
+    Map<ConfigResourceContext, Optional<ConfigDocument>> previousConfigDocs =
+        getLatestVersionConfigDocs(resourceContextValueMap.keySet());
+    Map<Key, Document> documentsToBeUpserted = new HashMap<>();
```
aaron-steinfeld, line 95: be careful with order here. We should be preserving order as far as we can, and that means using an ordered map to match the iteration order.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -76,6 +77,56 @@ public UpsertedConfig writeConfig(
...
+      Key latestDocKey = new ConfigDocumentKey(configResourceContext);
+      ConfigDocument latestConfigDocument =
+          buildConfigDocument(
+              configResourceContext,
+              resourceContextValueMap.get(configResourceContext),
+              userId,
+              previousConfigDoc);
+      documentsToBeUpserted.put(latestDocKey, latestConfigDocument);
+
+      upsertedConfigs.add(
```
aaron-steinfeld, line 111: this hasn't actually been upserted yet has it? Can we build the results after we upsert (would also allow us to switch this to a neater streaming implementation).

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -169,6 +209,43 @@ private Optional<ConfigDocument> getLatestVersionConfigDoc(
     return Optional.empty();
   }
 
+  private Map<ConfigResourceContext, Optional<ConfigDocument>> getLatestVersionConfigDocs(
+      Set<ConfigResourceContext> configResourceContexts) throws IOException {
+    // build filter
+    Filter[] childFilters = new Filter[configResourceContexts.size()];
```
aaron-steinfeld, line 215: If we're forced to use an array here (why did we use arrays in an API written this decade?!), we can still use a list and switch it over at the last second with toArray() so we don't ahve to deal with indices.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -169,6 +209,43 @@ private Optional<ConfigDocument> getLatestVersionConfigDoc(
...
+    int index = 0;
+    for (ConfigResourceContext configResourceContext : configResourceContexts) {
+      childFilters[index] = getConfigResourceContextFilter(configResourceContext);
+      index += 1;
+    }
+    Filter filter = new Filter();
+    filter.setOp(Filter.Op.OR);
```
aaron-steinfeld, line 222: Do we have IN filter support? Could probably do this as a single filter - here, we're making a pretty bulky tree - each one of these filters includes multiple AND'd filters that are duplicated for each context (tenant, resource, namespace).

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -169,6 +209,43 @@ private Optional<ConfigDocument> getLatestVersionConfigDoc(
...
+    filter.setChildFilters(childFilters);
+
+    // build query
+    Query query = new Query();
+    query.setFilter(filter);
+    query.addOrderBy(new OrderBy(VERSION_FIELD_NAME, false));
```
aaron-steinfeld, line 228: order isn't used in the response, why set it?

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -169,6 +209,43 @@ private Optional<ConfigDocument> getLatestVersionConfigDoc(
...
+    query.setLimit(configResourceContexts.size());
+
+    Map<ConfigResourceContext, Optional<ConfigDocument>> latestVersionConfigDocs = new HashMap<>();
```
aaron-steinfeld, line 231: Similar about order. The main point is that if a user passes in a set of update requests, they should be processed and returned in that order. The existing test that was forced to change should be a flashing red indicator that was broken.

config-service/build.gradle.kts
SrikarMannepalli: Right. Was looking to avoid `GrpcClientRequestContextUtil` in the newly added code as it is deprecated. Actually, let me replace all the occurrences as it isn't a large PR.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
SrikarMannepalli: I think we would only be duplicating tenant right? Because each of the configs to be upserted could have a different resource and namespace.
```
message UpsertAllConfigsRequest {
  repeated ConfigToUpsert configs = 1;

  message ConfigToUpsert {
    string resource_name = 1;
    string resource_namespace = 2;
    string context = 3;
    google.protobuf.Value config = 4;
  }
}
```

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -212,24 +219,22 @@ private Optional<ConfigDocument> getLatestVersionConfigDoc(
   private Map<ConfigResourceContext, Optional<ConfigDocument>> getLatestVersionConfigDocs(
       Set<ConfigResourceContext> configResourceContexts) throws IOException {
     // build filter
-    Filter[] childFilters = new Filter[configResourceContexts.size()];
-    int index = 0;
+    List<Filter> childFilters = new ArrayList<>();
     for (ConfigResourceContext configResourceContext : configResourceContexts) {
```
aaron-steinfeld, line 223: nit: now you can stream and map 😉

config-service/build.gradle.kts
aaron-steinfeld: I saw updates in tests and int tests, but no updates in the actual implementation. If I misread disregard, but if it's only being changed in tests can switch this dependency to be a testImplementation and integrationTestImplementation one.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -76,6 +79,61 @@ public UpsertedConfig writeConfig(
...
+  }
+
+  private List<UpsertedConfig> writeConfigs(
+      Map<ConfigResourceContext, Value> resourceContextValueMap, String userId) throws IOException {
+    Map<ConfigResourceContext, Optional<ConfigDocument>> previousConfigDocs =
+        getLatestVersionConfigDocs(resourceContextValueMap.keySet());
+    Map<Key, Document> documentsToBeUpserted = new LinkedHashMap<>();
+    for (Entry<ConfigResourceContext, Optional<ConfigDocument>> configResourceContextOptionalEntry :
+        previousConfigDocs.entrySet()) {
+      ConfigResourceContext configResourceContext = configResourceContextOptionalEntry.getKey();
```
aaron-steinfeld, line 100: nit: lots of nice iteration options on a map - like a foreach gives you a function accepting a key and value I believe.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -76,6 +79,61 @@ public UpsertedConfig writeConfig(
...
+      return zip(
+              previousConfigDocs.values().stream(),
+              documentsToBeUpserted.values().stream(),
+              (previousConfigDoc, documentToBeUpserted) ->
+                  previousConfigDoc
+                      .flatMap(this::convertToContextSpecificConfig)
+                      .map(
+                          previousConfig ->
+                              this.buildUpsertResult(
+                                  (ConfigDocument) documentToBeUpserted, previousConfig))
```
aaron-steinfeld, line 124: These casts are suspicious - is something mistyped? the zip should not lose the type of values in the stream.

SrikarMannepalli: `documentsToBeUpserted` isa `Map<Key, Document>` as `bulkUpsert` method expects it to be `Map<Key, Document>`. However, `buildUpsertResult` expects it to be a `ConfigDocument` because of which, having to cast it back.

aaron-steinfeld: So sounds like a mistype on bulkUpsert - ideally, we'd fix that to accept subtypes (not blocking though).

aaron-steinfeld (approved):

PR #149, opened 2023-01-04
feat: bulk delete configs support

SrikarMannepalli:
## Description
This PR adds the API contracts to support bulk delete of configs 

### Checklist:


config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -119,6 +121,31 @@ message DeleteConfigResponse {
...
+    // optional - only required if config applies to a specific context.
+    // If empty, specified config is associated with a default context.
+    string context = 3;
+  }
+}
+
+message DeleteConfigsResponse {
+  repeated DeletedConfig deleted_configs = 1;
+
+  message DeletedConfig {
```
aaron-steinfeld, line 141: The singular API returns a `ContextSpecificConfig` - why not reuse that?

SrikarMannepalli: I was contemplating between using it itself and a separate config, but went the way of `UpsertAllConfigs` finally. Updated it now.

skjindal93 (dismissed):

skjindal93 (dismissed):

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -140,6 +144,49 @@ public void deleteConfig(
...
+        return;
+      }
+      Map<ConfigResourceContext, Value> valuesByContext =
+          request.getConfigsList().stream()
+              .map(
+                  requestedDelete ->
+                      Map.entry(this.getConfigResourceContext(requestedDelete), emptyValue()))
+              .collect(ImmutableMap.toImmutableMap(Entry::getKey, Entry::getValue));
+
+      List<UpsertedConfig> deletedConfigs =
```
aaron-steinfeld, line 165: Wouldn't this be returning the post-delete values (aka empty). The response from this API I would expect would be the values that have been deleted, so if we're not going to return that we should be modifying the API.

aaron-steinfeld: I take it back - I see it returns the prev config too :)

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -140,6 +144,49 @@ public void deleteConfig(
...
+      responseObserver.onNext(
+          DeleteConfigsResponse.newBuilder()
+              .addAllDeletedConfigs(
+                  deletedConfigs.stream()
+                      .map(this::buildDeletedContextSpecificConfig)
+                      .collect(Collectors.toUnmodifiableList()))
+              .build());
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      log.error("Delete configs failed for request: {}", request, e);
```
aaron-steinfeld, line 177: responseObserver.onError missing here.

aaron-steinfeld (approved):

PR #151, opened 2023-01-05
fix: query filter bug

SrikarMannepalli:
## Description
This PR fixes a bug in the query filter. When we provide no configs to bulk update, the query filter default to an empty OR. This fetches all the configs throughout. We set a limit on the query response to 0 in such cases. But a limit of 0 on mongo is equivalent to no limit.  On update, all these are being set to null as it can't find the values to assign to it as they are not part of the request. 

### Testing

### Checklist:

### DIFF
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
@@ -208,14 +208,22 @@ public class DocumentConfigStore implements ConfigStore {
 
   private Map<ConfigResourceContext, Optional<ConfigDocument>> getLatestVersionConfigDocs(
       Set<ConfigResourceContext> configResourceContexts) throws IOException {
+    if (configResourceContexts.isEmpty()) {
+      return Collections.emptyMap();
+    }
     // build filter
     List<Filter> childFilters =
         configResourceContexts.stream()
-            .map(this::getConfigResourceContextFilter)
+            .map(this::getConfigResourceFieldContextFilter)
             .collect(Collectors.toUnmodifiableList());
-    Filter filter = new Filter();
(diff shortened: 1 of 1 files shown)


config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -208,14 +208,22 @@ private Optional<ConfigDocument> getLatestVersionConfigDoc(
 
   private Map<ConfigResourceContext, Optional<ConfigDocument>> getLatestVersionConfigDocs(
       Set<ConfigResourceContext> configResourceContexts) throws IOException {
+    if (configResourceContexts.isEmpty()) {
```
aaron-steinfeld, line 211: Should we be treating an updateAll with empty as valid? or rejecting the request?

aaron-steinfeld (approved):

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
SrikarMannepalli: Hmm🤔. We reject the request when we try to bulk upsert entities with empty request and also doc store `IN` operator does not take empty list as a valid request. Maybe here also we should reject the request? What do you think?

aaron-steinfeld: I would think we should be rejecting - it seems unintentional.

SrikarMannepalli: Sure. I've added this check with this PR - https://github.com/hypertrace/config-service/pull/152. Please review when possible.

PR #152, opened 2023-01-06
fix: disallow empty providing config list to bulk upsert 

SrikarMannepalli:
## Description
This PR adds a check to disallow providing empty config list to bulk upsert.

### Checklist:

### DIFF
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
@@ -144,6 +144,10 @@ public class ConfigServiceGrpcImpl extends ConfigServiceGrpc.ConfigServiceImplBa
   public void upsertAllConfigs(
       UpsertAllConfigsRequest request, StreamObserver<UpsertAllConfigsResponse> responseObserver) {
     try {
+      if (request.getConfigsCount() == 0) {
+        responseObserver.onError(Status.INVALID_ARGUMENT.asException());
+        return;
+      }
       Map<ConfigResourceContext, Value> valuesByContext =
           request.getConfigsList().stream()
               .map(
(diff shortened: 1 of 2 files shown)


config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -144,6 +144,10 @@ public void deleteConfig(
   public void upsertAllConfigs(
       UpsertAllConfigsRequest request, StreamObserver<UpsertAllConfigsResponse> responseObserver) {
     try {
+      if (request.getConfigsCount() == 0) {
+        responseObserver.onError(Status.INVALID_ARGUMENT.asException());
```
aaron-steinfeld, line 148: can you add a description on the error message

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -95,10 +96,14 @@ private List<UpsertedConfig> writeConfigs(
         getLatestVersionConfigDocs(resourceContextValueMap.keySet());
     Map<Key, Document> documentsToBeUpserted = new LinkedHashMap<>();
     previousConfigDocs.forEach(
-        (key, value) ->
-            documentsToBeUpserted.put(
-                new ConfigDocumentKey(key),
-                buildConfigDocument(key, resourceContextValueMap.get(key), userId, value)));
+        (key, value) -> {
+          if (!resourceContextValueMap.containsKey(key)) {
+            throw Status.INVALID_ARGUMENT.asRuntimeException();
```
aaron-steinfeld, line 101: this shouldn't happen right? If so, please use an internal error

aaron-steinfeld (approved):

PR #153, opened 2023-01-11
chore: bulk delete configs in Identified object store

SrikarMannepalli:
## Description
This PR adds bulk delete configs method in Identified object store

### Testing
I've added unit tests to validate the changes
### Checklist:


object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
```diff
@@ -205,6 +198,29 @@ public List<ContextualConfigObject<T>> upsertObjects(RequestContext context, Lis
...
+                        .build())
+            .collect(Collectors.toUnmodifiableList());
+    return requestContext
+        .call(
+            () ->
+                this.configServiceBlockingStub.deleteConfigs(
+                    DeleteConfigsRequest.newBuilder().addAllConfigs(configsToDelete).build()))
+        .getDeletedConfigsList()
+        .stream()
+        .map(deletedConfig -> processDeleteResult(requestContext, deletedConfig))
```
aaron-steinfeld, line 220: If I delete 10 things and the first fails to process, no notifications will be sent for any of them. We should probably wait to throw until the end - if at all.

SrikarMannepalli: Right, okay. Also should we throw an error or just log one and continue?🤔 What do you suggest?

aaron-steinfeld: So thinking through it, what's likely to fail in this function? The conversion from value to proto right? If that happens, from a delete API perspective, it's not _really_ an error - it deleted just fine (I'd call this an issue with existing single delete too).

So I would probably process a result by gathering the ID that was deleted and an optional config if we can parse it. 

On the method return (both singular and bulk), gather all the deleted IDs regardless of whether we parsed the configs or not and return those. Basically method should only error if it actually didn't delete.

On the config change event - on the read side, does it support leaving the prev value unset and just setting the ID? If so, seems better than nothing. If not, skip sending it. Either way, log a warning for that, but still return the id to the delete result. 

If

SrikarMannepalli: Skipping sending it for now.
We probably can set the prev value to some null value, but would need to send the config type as part of the delete notification which we can't parse unless we can successfully convert the value to proto. I'm not sure if there would be any use if we send the delete notification setting the config type to say an empty string. Also, not sure if they would get parsed with the current consumer side implementation. What do you think? 🤔

aaron-steinfeld: (so yes, agree to skip sending for now)

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
```diff
@@ -238,6 +255,33 @@ private Optional<ContextualConfigObject<T>> processUpsertResult(
...
+              configChangeEventGenerator.sendDeleteNotification(
+                  requestContext,
+                  this.buildClassNameForChangeEvent(contextualConfigObject.getData()),
+                  deletedConfig.getContext(),
+                  this.buildValueForChangeEvent(contextualConfigObject.getData())));
+    } else {
+      contextualConfigObject =
+          new ContextualConfigObjectImpl<>(
+              deletedConfig.getContext(),
+              null,
```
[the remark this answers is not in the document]
SrikarMannepalli: I've verified that there are consumers which rely on the data field being non-null. Didn't quite get the part regarding subtype 😅. Could you please elaborate on that?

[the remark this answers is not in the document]
aaron-steinfeld: > I've verified that there are consumers which rely on the data field being non-null.

OK so sounds like we need another option. What I meant with a subtype was something like 

``` class DeletedContextualConfigObject extends ContextualConfigObjectImpl {
// Change constructor to not require a config
```

This helps in this spot, but at the end of the day either way - we need to make sure getConfig() returns something for those existing consumers. Maybe that's an empty config? The subclass doesn't really buy us much there though - so if that's what we're doing, I wouldn't bother. the subclass would be useful if we want to put some logic into a getConfig call or, for example, deprecate the returned config on delete only.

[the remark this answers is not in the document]
SrikarMannepalli: Hmm, got it. One solution I see is that we add a new abstract method which builds a default instance of the proto message(basically a default/empty config as you had mentioned). But this would involve adding the method in all the places where `IdentifiedObjectStore` is used. What do you think? 🤔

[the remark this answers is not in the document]
SrikarMannepalli: Got it. I've tried to make the changes as per 1. But they don't look too clean 😅. What do you think? 🤔

object-store/src/main/java/org/hypertrace/config/objectstore/ConfigObject.java
```diff
@@ -1,9 +1,10 @@
 package org.hypertrace.config.objectstore;
 
 import java.time.Instant;
+import java.util.Optional;
 
 public interface ConfigObject<T> {
-  T getData();
+  Optional<T> getData();
```
aaron-steinfeld, line 7: I think there was a misunderstanding here - getData should not be made optional. _Delete's_ return can be made optional. In every scenario beside delete, we don't want this to be empty, so now we've just made dozens of usages in this PR alone have to call `.get()` on an optional, which is a big code smell. If we have to call `.get()`, we've messed up.

My suggestion had been to change the object stores so instead of
```
public Optional<ContextualConfigObject<T>> deleteObject(
```

return something like
```
public DeleteResult<T> deleteObject(
```
Or potentially (this one probably will make it cleaner, since most result handling just works with the optional today)
```
public Optional<DeletedContextualConfigObject<T>> deleteObject(
```

In this way, we're only breaking consumers of the delete API and in the latter one, only that subset of consumers that then needs to actually read data from the result.

SrikarMannepalli: Right, got it. Updated it now.

object-store/src/main/java/org/hypertrace/config/objectstore/DeletedConfigObject.java
```diff
@@ -0,0 +1,12 @@
+package org.hypertrace.config.objectstore;
+
+import java.time.Instant;
+import java.util.Optional;
+
+public interface DeletedConfigObject<T> {
+  Optional<T> getOptionalData();
```
aaron-steinfeld, line 7: nit -
```suggestion
  Optional<T> getDeletedData();
```

aaron-steinfeld (approved):

aaron-steinfeld (approved):

PR #154, opened 2023-01-21
Add partitioner config service 

surajpuvvada:
1. Introduce GlobalConfigServiceFactory to create global services that are not tied to a tenant
2. Add TenantIsolationConfigService that managed tenant isolation config. This config will be fetched by all stream processing kafka streams apps


config-service/src/integrationTest/java/org/hypertrace/config/service/ConfigServiceIntegrationTest.java
```diff
@@ -1,97 +1,20 @@
...
-import org.hypertrace.config.service.v1.UpsertAllConfigsRequest;
-import org.hypertrace.config.service.v1.UpsertAllConfigsRequest.ConfigToUpsert;
-import org.hypertrace.config.service.v1.UpsertAllConfigsResponse;
-import org.hypertrace.config.service.v1.UpsertConfigRequest;
-import org.hypertrace.config.service.v1.UpsertConfigResponse;
-import org.hypertrace.core.documentstore.Collection;
-import org.hypertrace.core.documentstore.Datastore;
-import org.hypertrace.core.documentstore.DatastoreProvider;
-import org.hypertrace.core.grpcutils.client.RequestContextClientCallCredsProviderFactory;
+import org.hypertrace.config.service.v1.*;
```
aaron-steinfeld, line 13: Fix imports

config-service/src/integrationTest/java/org/hypertrace/config/service/ConfigServiceIntegrationTest.java
```diff
@@ -1,97 +1,20 @@
...
-  public static void teardown() {
-    IntegrationTestServerUtil.shutdownServices();
-  }
-
-  // Need to delete the collection after each test for stateless integration testing
-  @AfterEach
-  void delete() {
-    CONFIGURATIONS_COLLECTION.deleteAll();
-  }
+public class ConfigServiceIntegrationTest extends ConfigServiceIntegrationTestBase {
```
aaron-steinfeld, line 17: Test base classes are an anti-pattern and an abuse of inheritance in order to re-use code without an is-a relationship. They exist in the code in some places, but let's not go out of our way to introduce them. Prefer composition.

config-service/src/integrationTest/resources/configs/config-service/application.conf
```diff
@@ -2,6 +2,7 @@ main.class = org.hypertrace.config.service.ConfigService
 service.name = config-service
 service.port = 50101
 service.admin.port = 50102
+global.config.service.port.internal = 50103
```
aaron-steinfeld, line 5: what's internal mean here? Also, nothing needs to be declared in here unless it's overriding the defaults - and this looks to be a duplicate

tenant-isolation-config-service-api/src/main/proto/buf.yaml
```diff
@@ -0,0 +1,8 @@
+version: v1
```
aaron-steinfeld, line 1: make sure to declare this new source root in the top level work yaml.

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
```diff
@@ -0,0 +1,44 @@
...
+import "org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_group_config.proto";
+
+service TenantIsolationConfigService{
+  rpc GetAllTenantIsolationGroupConfigs(GetAllTenantIsolationGroupConfigsRequest) returns (GetAllTenantIsolationGroupConfigsResponse) {}
+  rpc CreateTenantIsolationGroupConfig(CreateTenantIsolationGroupConfigRequest) returns (CreateTenantIsolationGroupConfigResponse) {}
+  rpc UpdateTenantIsolationGroupConfig(UpdateTenantIsolationGroupConfigRequest) returns (UpdateTenantIsolationGroupConfigResponse) {}
+  rpc DeleteTenantIsolationGroupConfig(DeleteTenantIsolationGroupConfigRequest) returns (DeleteTenantIsolationGroupConfigResponse) {}
+}
+
+message GetAllTenantIsolationGroupConfigsRequest{
```
aaron-steinfeld, line 16: check style - should be a space before brace at least

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
```diff
@@ -0,0 +1,44 @@
+syntax = "proto3";
+
+package org.hypertrace.tenantisolation.config.service.v1;
+
+option java_multiple_files = true;
+
+import "org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_group_config.proto";
+
+service TenantIsolationConfigService{
```
aaron-steinfeld, line 9: The naming here feels a bit off (applies to whole PR). It's not really _isolation_ so much as defining how to partition tenants which may or may not be isolated, right?

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
```diff
@@ -0,0 +1,44 @@
...
+  rpc GetAllTenantIsolationGroupConfigs(GetAllTenantIsolationGroupConfigsRequest) returns (GetAllTenantIsolationGroupConfigsResponse) {}
```
aaron-steinfeld, line 10: Instead of `GetAll`, just do a plural Get. That can model get all, but it also allows filtered queries if needed (so is easier to evolve over time).

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
```diff
@@ -0,0 +1,44 @@
...
+  rpc CreateTenantIsolationGroupConfig(CreateTenantIsolationGroupConfigRequest) returns (CreateTenantIsolationGroupConfigResponse) {}
```
aaron-steinfeld, line 11: In our previous conversations, you had mentioned that only read was needed at the moment. If that's still the case, can we drop the write side of this PR until and unless needed? If not, what's changed?

This PR is already far too big to meaningfully review -  and if we're not using write at the moment, we're just guessing as to the requirements there which often leads to breaking changes.

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
```diff
@@ -0,0 +1,44 @@
...
+message CreateTenantIsolationGroupConfigRequest{
+  NewTenantIsolationGroupConfig config = 1;
+}
+
+message CreateTenantIsolationGroupConfigResponse{
+  TenantIsolationGroupConfig config = 1;
+}
+
+message UpdateTenantIsolationGroupConfigRequest{
+  TenantIsolationGroupConfig config = 1;
```
aaron-steinfeld, line 32: accept a different item for update than you return. The few lines of code we save by reusing a type is not worth forever coupling the request and response. It would prevent us from ever representing read-only fields, for example.

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
```diff
@@ -0,0 +1,44 @@
...
+}
+
+message UpdateTenantIsolationGroupConfigResponse{
+  TenantIsolationGroupConfig config = 1;
+}
+
+message DeleteTenantIsolationGroupConfigRequest{
```
aaron-steinfeld, line 39: Delete should work in IDs.

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_group_config.proto
```diff
@@ -0,0 +1,18 @@
+syntax = "proto3";
+
+package org.hypertrace.tenantisolation.config.service.v1;
+
+option java_multiple_files = true;
+
+message TenantIsolationGroupConfig{
+  string id = 1;
+  string group_name = 2;
```
aaron-steinfeld, line 9: What's the group name used for?

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_group_config.proto
```diff
@@ -0,0 +1,18 @@
...
+  repeated string members = 3;
```
aaron-steinfeld, line 10: be specific - what does a member string represent? Guessing it should be `member_tenant_ids`?

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+import org.hypertrace.tenantisolation.config.service.store.TenantIsolationConfigStoreProvider;
+import org.hypertrace.tenantisolation.config.service.store.TenantIsolationGroupConfigDTO;
+import org.hypertrace.tenantisolation.config.service.v1.*;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class TenantIsolationConfigServiceImpl
+    extends TenantIsolationConfigServiceGrpc.TenantIsolationConfigServiceImplBase {
+
+  private static final Logger logger =
```
aaron-steinfeld, line 18: you're using lombok already, so to be consistent use `@Slf4j`

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+      LoggerFactory.getLogger(TenantIsolationConfigServiceImpl.class);
+
+  private static final String GENERIC_CONFIG_SERVICE = "generic.config.service";
+
+  private TenantIsolationConfigServiceRequestValidator validator =
+      new TenantIsolationConfigServiceRequestValidator();
```
aaron-steinfeld, line 24: When using dependency injection, inject your dependencies rather than manually constructing them.

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+
+  private TenantIsolationConfigStore tenantIsolationConfigStore;
+
+  @Inject
+  public TenantIsolationConfigServiceImpl(Config config) {
+    Config cfg = config.getConfig(GENERIC_CONFIG_SERVICE);
+    this.tenantIsolationConfigStore = TenantIsolationConfigStoreProvider.getDocumentStore(cfg);
```
aaron-steinfeld, line 31: Prefer injecting this especially - makes testing much nicer to be able to mock it.

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  public void getAllTenantIsolationGroupConfigs(
+      GetAllTenantIsolationGroupConfigsRequest request,
+      StreamObserver<GetAllTenantIsolationGroupConfigsResponse> responseObserver) {
+    try {
+      validator.validateOrThrow(request);
+      List<TenantIsolationGroupConfigDTO> configDTOs =
```
aaron-steinfeld, line 39: camelCase - `Dto`

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+                      .setGroupName(upsertedConfigDTO.getGroupName())
+                      .setWeight(upsertedConfigDTO.getWeight())
+                      .addAllMembers(upsertedConfigDTO.getMembers())
+                      .build())
+              .build());
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      logger.error(
+          String.format(
+              "createTenantIsolationGroupConfig failed for request {%s} with error ", request),
```
aaron-steinfeld, line 86: logger already has a built in template engine, no need to repeat

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+
+  public void createTenantIsolationGroupConfig(
+      CreateTenantIsolationGroupConfigRequest request,
+      StreamObserver<CreateTenantIsolationGroupConfigResponse> responseObserver) {
+    try {
+      validator.validateOrThrow(request);
+      TenantIsolationGroupConfigDTO upsertedConfigDTO =
+          tenantIsolationConfigStore.upsert(
```
aaron-steinfeld, line 67: What's the purpose of this DTO? Seems like complexity for no benefit. Given we have equivalent compatibility requirements on both API and storage, generally we use the same item. Even if they ever were to diverge in the future, we can still do that just by diverging the implementation.

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+    } catch (Exception e) {
+      logger.error(
+          String.format(
+              "createTenantIsolationGroupConfig failed for request {%s} with error ", request),
+          e);
+      responseObserver.onError(e);
+    }
+  }
+
+  /** */
```
aaron-steinfeld, line 92: nit - remove or fill in doc

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,56 @@
...
+  public void validateOrThrow(DeleteTenantIsolationGroupConfigRequest request) {
+    if (request.getGroupName().isBlank()) {
+      throw Status.INVALID_ARGUMENT
+          .withDescription("groupName can't be empty")
+          .asRuntimeException();
+    }
+  }
+
+  private void validateTenantIsolationConfigOrThrow(TenantIsolationGroupConfig config) {
+    if (config == null) {
```
aaron-steinfeld, line 28: not reachable - protos never use null.

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,56 @@
...
+      throw Status.INVALID_ARGUMENT.withDescription("Config can't be empty").asRuntimeException();
+    }
+    if (config.getGroupName().isBlank()) {
+      throw Status.INVALID_ARGUMENT
+          .withDescription("groupName can't be empty")
+          .asRuntimeException();
+    }
+
+    if (config.getMembersList().isEmpty()) {
+      throw Status.INVALID_ARGUMENT.withDescription("members can't be empty").asRuntimeException();
```
aaron-steinfeld, line 38: What about the default for an unknown tenant? how is that represented?

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
```diff
@@ -0,0 +1,141 @@
...
+              TenantIsolationGroupConfigDTO.builder()
+                  .groupName(request.getConfig().getGroupName())
+                  .weight(request.getConfig().getWeight())
+                  .members(request.getConfig().getMembersList())
+                  .build());
+      responseObserver.onNext(
+          CreateTenantIsolationGroupConfigResponse.newBuilder()
+              .setConfig(
+                  TenantIsolationGroupConfig.newBuilder()
+                      .setGroupName(upsertedConfigDTO.getGroupName())
```
aaron-steinfeld, line 77: ID is never being set in either direction.

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/store/TenantIsolationConfigDocumentStore.java
```diff
@@ -0,0 +1,45 @@
+package org.hypertrace.tenantisolation.config.service.store;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
+import org.hypertrace.core.documentstore.*;
+
+public class TenantIsolationConfigDocumentStore implements TenantIsolationConfigStore {
```
aaron-steinfeld, line 8: Thought you had mentioned that it would just be using static config for now?

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/store/TenantIsolationConfigStoreProvider.java
```diff
@@ -0,0 +1,20 @@
...
+import org.hypertrace.core.documentstore.Datastore;
+import org.hypertrace.core.documentstore.DatastoreProvider;
+
+public class TenantIsolationConfigStoreProvider {
+  public static final String DOC_STORE_CONFIG_KEY = "document.store";
+  public static final String DATA_STORE_TYPE = "dataStoreType";
+
+  private TenantIsolationConfigStoreProvider() {}
+
+  public static TenantIsolationConfigDocumentStore getDocumentStore(Config config) {
```
aaron-steinfeld, line 13: no statics - use DI consistently

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/store/TenantIsolationGroupConfigDocument.java
```diff
@@ -0,0 +1,38 @@
+package org.hypertrace.tenantisolation.config.service.store;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.DeserializationFeature;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.hypertrace.core.documentstore.Document;
+
+public class TenantIsolationGroupConfigDocument implements Document {
+
+  private static final ObjectMapper OBJECT_MAPPER =
```
aaron-steinfeld, line 10: if we had just read/written the proto, we wouldn't have to manually map this either.

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/store/TenantIsolationGroupConfigKey.java
```diff
@@ -0,0 +1,15 @@
...
+import lombok.Value;
+import org.hypertrace.core.documentstore.Key;
+
+@Value
+public class TenantIsolationGroupConfigKey implements Key {
+  String groupName;
+
+  @Override
+  public String toString() {
+    return UUID.nameUUIDFromBytes((groupName).getBytes()).toString();
```
aaron-steinfeld, line 13: Why do this on each read rather than assigning an ID? Both requires multiple executions and is just generally more confusing.

config-service/src/integrationTest/resources/configs/config-service/application.conf
surajpuvvada: I was following the pattern I'm guessing it means it is an internal port and not exposed externally. Will remove it from here.

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/store/TenantIsolationConfigDocumentStore.java
surajpuvvada: Not sure I understood your question ? We will need to support read/write. The write will be via a separate tool which SREs would use to generate new config and update. Do you have any concerns ?

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
surajpuvvada: yea not very happy with the naming here - how does TenantPartitionGroupConfigService sound ?

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
surajpuvvada: We will need write too. Just that it won't be through any platform service but via a tool that SREs will use. For example if it is a cli it could still use the grpc APIs to update the config

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
surajpuvvada: Is there some example of how the _id field (mongo object field) is mapped ?

config-service/src/integrationTest/resources/configs/config-service/application.conf
aaron-steinfeld: whether it's exposed or not wouldn't be decided in this config - although we do have the concept of config services meant for inside the platform and those meant for external consumers - maybe that's what you saw?

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
aaron-steinfeld: I'd probably go even simpler to just `TenantPartitioningConfigService`

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
aaron-steinfeld: There's a use case for write, yes - my point was that it didn't need to be all in one giant first cut. The requirements you described for the original use case don't need write, and that would have greatly simplified the PR - that's why I thought we had discussed just populating this via helm config.

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/store/TenantIsolationConfigDocumentStore.java
aaron-steinfeld: I don't have concerns with supporting write in general, just with 1. dropping the whole read/write/deploy in a single giant PR all at once and 2. building an API before we're ready to consume it

config-service/src/integrationTest/resources/configs/config-service/application.conf
surajpuvvada: removed it from this file

config-service/src/main/resources/configs/common/application.conf
```diff
@@ -2,6 +2,8 @@ main.class = org.hypertrace.config.service.ConfigService
 service.name = config-service
 service.port = 50101
 service.admin.port = 50102
+global.config.service.port.internal = 50103
```
surajpuvvada, line 5: I can remove the internal suffix if you prefer - but this was meant to convey it is an internal port

aaron-steinfeld: all of the other ones above it are internal too though - removing it would be my preference.

tenant-partitioning-config-service-api/src/main/proto/org/hypertrace/tenantpartitioning/config/service/v1/tenant_partition_group_config.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package org.hypertrace.tenantpartitioning.config.service.v1;
+
+option java_multiple_files = true;
+
+message TenantPartitionGroupConfig{
+  string id = 1;
+  string group_name = 2;
```
aaron-steinfeld, line 9: what do we do with name? just for logging?

tenant-partitioning-config-service-api/src/main/proto/org/hypertrace/tenantpartitioning/config/service/v1/tenant_partition_group_config.proto
```diff
@@ -0,0 +1,23 @@
...
+  int32 weight = 4;
+}
+
+message NewTenantPartitionGroupConfig{
+  string group_name = 1;
+  repeated string member_tenant_ids = 2;
+  int32 weight = 3;
+}
+
+message TenantPartitionGroupConfigForUpdate{
```
aaron-steinfeld, line 20: how's this work without an id?

tenant-partitioning-config-service-impl/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
+plugins {
+  `java-library`
+  jacoco
+  id("org.hypertrace.jacoco-report-plugin")
+}
+
+dependencies {
+  api(projects.tenantPartitioningConfigServiceApi)
+  implementation(projects.configServiceApi)
```
aaron-steinfeld, line 9: looks like there's likely some copy paste in this dependency list. This is independent of the config service right? Please remove anything not being used.

tenant-partitioning-config-service-impl/src/main/java/org/hypertrace/tenantpartitioning/config/service/TenantPartitioningConfigServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+
+  @Inject
+  public TenantPartitioningConfigServiceImpl(
+      Config config,
+      TenantPartitionConfigStore tenantIsolationConfigStore,
+      TenantPartitioningConfigServiceRequestValidator validator) {
+    this.tenantIsolationConfigStore = tenantIsolationConfigStore;
+    this.validator = validator;
+    Config cfg = config.getConfig(GENERIC_CONFIG_SERVICE);
+    // this.tenantIsolationConfigStore = TenantPartitionConfigStoreProvider.getDocumentStore(cfg);
```
aaron-steinfeld, line 30: remove commented out code.

tenant-partitioning-config-service-impl/src/main/java/org/hypertrace/tenantpartitioning/config/service/TenantPartitioningConfigServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+  private final TenantPartitioningConfigServiceRequestValidator validator;
+
+  @Inject
+  public TenantPartitioningConfigServiceImpl(
+      Config config,
+      TenantPartitionConfigStore tenantIsolationConfigStore,
+      TenantPartitioningConfigServiceRequestValidator validator) {
+    this.tenantIsolationConfigStore = tenantIsolationConfigStore;
+    this.validator = validator;
+    Config cfg = config.getConfig(GENERIC_CONFIG_SERVICE);
```
aaron-steinfeld, line 29: Prefer using a typed class to parse the untyped config before handing it to the injector. That way, the config structure is kept decoupled from code and any config issues cause a startup failure rather than failing on the first request that tries to access it.

tenant-partitioning-config-service-impl/src/main/java/org/hypertrace/tenantpartitioning/config/service/TenantPartitioningConfigServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+    // this.tenantIsolationConfigStore = TenantPartitionConfigStoreProvider.getDocumentStore(cfg);
+  }
+
+  public void getTenantPartitionGroupConfigs(
+      GetTenantPartitionGroupConfigsRequest request,
+      StreamObserver<GetTenantPartitionGroupConfigsResponse> responseObserver) {
+    try {
+      validator.validateOrThrow(request);
+      List<TenantPartitionGroupConfig> configDTOs = tenantIsolationConfigStore.getAllGroupConfigs();
```
aaron-steinfeld, line 38: these aren't DTOs any more - please fix naming. Also we seem to stream and map these into the same type - can that be removed?

tenant-partitioning-config-service-impl/src/main/java/org/hypertrace/tenantpartitioning/config/service/TenantPartitioningConfigServiceImpl.java
```diff
@@ -0,0 +1,140 @@
...
+
+  public void createTenantPartitionGroupConfig(
+      CreateTenantPartitionGroupConfigRequest request,
+      StreamObserver<CreateTenantPartitionGroupConfigResponse> responseObserver) {
+    try {
+      validator.validateOrThrow(request);
+      TenantPartitionGroupConfig upsertedConfigDTO =
+          tenantIsolationConfigStore.upsert(
+              TenantPartitionGroupConfig.newBuilder()
+                  .setId(
```
aaron-steinfeld, line 68: If we're not going to use uuids for identification, make sure to validate that the ID handed to create doesn't already exist.

For that matter, probably need to validate (on both create and update) that the tenant ids provided do not belong to another group.

tenant-partitioning-config-service-impl/src/main/java/org/hypertrace/tenantpartitioning/config/service/store/TenantPartitionConfigStoreProvider.java
```diff
@@ -0,0 +1,20 @@
...
+import org.hypertrace.core.documentstore.Datastore;
+import org.hypertrace.core.documentstore.DatastoreProvider;
+
+public class TenantPartitionConfigStoreProvider {
+  public static final String DOC_STORE_CONFIG_KEY = "document.store";
+  public static final String DATA_STORE_TYPE = "dataStoreType";
+
+  private TenantPartitionConfigStoreProvider() {}
+
+  public static TenantPartitionConfigDocumentStore getDocumentStore(Config config) {
```
aaron-steinfeld, line 13: nit - prefer not static, use via guice.

config-service-factory/src/main/java/org/hypertrace/config/service/GlobalConfigServiceFactory.java
```diff
@@ -0,0 +1,26 @@
+package org.hypertrace.config.service;
```
surajpuvvada, line 1: will move this to TraceableConfigService

config-service/src/integrationTest/java/org/hypertrace/config/service/tenantpartitioning/TenantPartitioningConfigServiceIntegrationTest.java
```diff
@@ -0,0 +1,298 @@
...
+import org.hypertrace.tenantpartitioning.config.service.v1.PutTenantPartitionGroupsConfigResponse;
+import org.hypertrace.tenantpartitioning.config.service.v1.TenantPartitionGroupConfig;
+import org.hypertrace.tenantpartitioning.config.service.v1.TenantPartitionGroupsConfig;
+import org.hypertrace.tenantpartitioning.config.service.v1.TenantPartitioningConfigServiceGrpc;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class TenantPartitioningConfigServiceIntegrationTest {
```
surajpuvvada, line 32: will move this to TraceableConfigService

config-service/src/main/resources/configs/common/application.conf
```diff
@@ -2,6 +2,7 @@ main.class = org.hypertrace.config.service.ConfigService
 service.name = config-service
 service.port = 50101
 service.admin.port = 50102
+global.config.service.port = 50103
```
surajpuvvada, line 5: will move this to TraceableConfigService

config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
```diff
@@ -12,6 +12,8 @@
 public class ConfigService extends StandAloneGrpcPlatformServiceContainer {
 
   private final ConfigServiceFactory configServiceFactory = new ConfigServiceFactory();
+  private final GlobalConfigServiceFactory globalConfigServiceFactory =
```
surajpuvvada, line 15: will move this to TraceableConfigService

tenant-partitioning-config-service-api/src/main/proto/buf.yaml
```diff
@@ -0,0 +1,8 @@
+version: v1
```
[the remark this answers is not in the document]
surajpuvvada: done

tenant-partitioning-config-service-api/src/main/proto/org/hypertrace/tenantpartitioning/config/service/v1/tenant_partition_group_config.proto
surajpuvvada: most informational purpose

config-service/src/integrationTest/java/org/hypertrace/config/service/ConfigServiceIntegrationTest.java
surajpuvvada: removed

tenant-isolation-config-service-api/src/main/proto/buf.yaml
surajpuvvada: done

tenant-isolation-config-service-api/src/main/proto/org/hypertrace/tenantisolation/config/service/v1/tenant_isolation_config_service.proto
surajpuvvada: fixed

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/TenantIsolationConfigServiceImpl.java
surajpuvvada: fixed

tenant-isolation-config-service-impl/src/main/java/org/hypertrace/tenantisolation/config/service/store/TenantIsolationGroupConfigDocument.java
surajpuvvada: fixed by directly using proto <-> json

config-service/src/main/resources/configs/common/application.conf
surajpuvvada: done

tenant-partitioning-config-service-api/src/main/proto/org/hypertrace/tenantpartitioning/config/service/v1/tenant_partition_group_config.proto
surajpuvvada: removed

tenant-partitioning-config-service-impl/build.gradle.kts
surajpuvvada: done

surajpuvvada:
can't see why validate-protos is failing - any ideas @aaron-steinfeld ? Could it be due to the renaming and deletions in proto files ?

config-service-factory/build.gradle.kts
```diff
@@ -15,4 +15,5 @@ dependencies {
   implementation(projects.notificationRuleConfigServiceImpl)
   implementation(projects.notificationChannelConfigServiceImpl)
   implementation(projects.spanProcessingConfigServiceImpl)
+  implementation(projects.tenantPartitioningConfigServiceImpl)
```
aaron-steinfeld, line 18: changes can be reverted

tenant-partitioning-config-service-api/src/main/proto/org/hypertrace/tenantpartitioning/config/service/v1/tenant_partition_group_config.proto
```diff
@@ -0,0 +1,23 @@
+syntax = "proto3";
+
+package org.hypertrace.tenantpartitioning.config.service.v1;
+
+option java_multiple_files = true;
+
+message TenantPartitionGroupConfig{
+  string id = 1;
+  string group_name = 2;
+  repeated string member_tenant_ids = 3;
```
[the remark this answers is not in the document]
aaron-steinfeld: > so the group names are used to encode the context.

So this sounds like why group name is necessary - there's an implicit meaning to certain group names that invalidates certain fields in the config? That's bad API design and easily fixed. Instead, we can do something like


```proto
message TenantPartitionGroupsConfig {
  repeated TenantPartitionGroup groups = 1;
  DefaultTenantPartitionGroup default_group = 2;
}

message TenantPartitionGroup {
  repeated string member_tenant_ids = 1;
  int32 weight = 2;
}

message DefaultTenantPartitionGroup {
  int32 weight = 1;
}
```

Not only does this more accurately display what's expected/required, but it makes it significantly easier to do validations like "there must be a default group", and "every other group must have tenant ids".

tenant-partitioning-config-service-impl/src/main/java/org/hypertrace/tenantpartitioning/config/service/TenantPartitioningConfigServiceImpl.java
```diff
@@ -0,0 +1,78 @@
+package org.hypertrace.tenantpartitioning.config.service;
+
+import com.google.inject.Inject;
+import io.grpc.stub.StreamObserver;
+import java.util.Optional;
+import lombok.extern.slf4j.Slf4j;
+import org.hypertrace.tenantpartitioning.config.service.store.TenantPartitionGroupsConfigStore;
+import org.hypertrace.tenantpartitioning.config.service.v1.*;
```
aaron-steinfeld, line 8: please fix imports (and IDE)

tenant-partitioning-config-service-impl/src/main/java/org/hypertrace/tenantpartitioning/config/service/store/TenantPartitionGroupsConfigDocumentStore.java
```diff
@@ -0,0 +1,42 @@
...
+  public TenantPartitionGroupsConfigDocumentStore(Datastore datastore) {
+    this.collection = datastore.getCollection(TENANT_ISOLATION_CONFIG_COLLECTION);
+  }
+
+  @Override
+  public Optional<TenantPartitionGroupsConfig> getConfig() throws IOException {
+    try (CloseableIterator<Document> it = collection.search(new Query())) {
+      if (it.hasNext()) {
+        String json = it.next().toJson();
+        return Optional.of(TenantPartitionGroupsConfigDocument.fromJson(json));
```
aaron-steinfeld, line 23: weird to return mid iterator - can we collect all of it and then error if more than one else return.

aaron-steinfeld:
> can't see why validate-protos is failing - any ideas @aaron-steinfeld ? Could it be due to the renaming and deletions in proto files ?

known bug - when a new root is added, it's incorrectly detected as a breaking change. As long as that's only issue, we'll override.

tenant-partitioning-config-service-impl/src/main/java/org/hypertrace/tenantpartitioning/config/service/store/TenantPartitionGroupsConfigDocumentStore.java
```diff
@@ -1,26 +1,31 @@
...
   @Override
   public Optional<TenantPartitionGroupsConfig> getConfig() throws IOException {
     try (CloseableIterator<Document> it = collection.search(new Query())) {
-      if (it.hasNext()) {
-        String json = it.next().toJson();
-        return Optional.of(TenantPartitionGroupsConfigDocument.fromJson(json));
+      List<Document> list = new ArrayList<>();
+      while (it.hasNext()) {
+        list.add(it.next());
+      }
```
aaron-steinfeld, line 24:
```suggestion
      List<Document> documents = ImmutableList.copyOf(it);
```

config-service/build.gradle.kts
```diff
@@ -70,6 +70,7 @@ dependencies {
   integrationTestImplementation(libs.hypertrace.grpcutils.client)
   integrationTestImplementation(libs.hypertrace.grpcutils.context)
   integrationTestImplementation(libs.hypertrace.documentstore)
+  integrationTestImplementation(projects.partitionerConfigServiceImpl)
```
surajpuvvada, line 73: will remove after code review

config-service/src/integrationTest/java/org/hypertrace/config/service/partitioner/PartitionerConfigServiceIntegrationTest.java
```diff
@@ -0,0 +1,289 @@
...
+import org.hypertrace.core.grpcutils.client.RequestContextClientCallCredsProviderFactory;
+import org.hypertrace.core.serviceframework.IntegrationTestServerUtil;
+import org.hypertrace.partitioner.config.service.store.PartitionerProfilesDocumentStore;
+import org.hypertrace.partitioner.config.service.v1.*;
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+public class PartitionerConfigServiceIntegrationTest {
```
surajpuvvada, line 23: will remove after code review

config-service/src/main/java/org/hypertrace/config/service/ConfigService.java
```diff
@@ -31,6 +33,12 @@ protected List<GrpcPlatformServerDefinition> getServerDefinitions() {
             .name(this.getServiceName())
             .port(this.getServicePort())
             .serviceFactory(this.configServiceFactory)
+            .build(),
+        GrpcPlatformServerDefinition.builder()
```
surajpuvvada, line 37: will remove after code review

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
```diff
@@ -0,0 +1,60 @@
...
+option java_multiple_files = true;
+
+
+service PartitionerConfigService {
+
+  rpc GetPartitionerProfile(GetPartitionerProfileRequest) returns (GetPartitionerProfileResponse) {}
+
+  rpc PutPartitionerConfig(PutPartitionerConfigRequest) returns (PutPartitionerConfigResponse) {}
+
+  rpc GetAllPartitionerConfig(GetAllPartitionerConfigRequest) returns (GetAllPartitionerConfigResponse) {}
```
aaron-steinfeld, line 14: micro-nit: put get and get all next to each other

aaron-steinfeld: also config was replaced with profiles. GetAll, let's change to plural get - that way, if we ever need to add the ability to filter in the future it slots in cleanly.

```suggestion
  rpc GetPartionerProfiles(GetPartionerProfilesRequest) returns (GetPartionerProfilesResponse) {}
```

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
```diff
@@ -0,0 +1,60 @@
...
+
+  rpc DeletePartitionerConfig(DeletePartitionerConfigRequest) returns (DeletePartitionerConfigResponse) {}
+
+}
+
+message GetPartitionerProfileRequest {
+  string profile = 1; // not optional
```
aaron-steinfeld, line 21:
```suggestion
  string profile_name = 1;
```

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
```diff
@@ -0,0 +1,60 @@
...
+}
+
+message GetPartitionerProfileResponse {
+  PartitionerProfile profile = 1;
+}
+
+message PutPartitionerConfigRequest {
+  PartitionerProfile profile = 1;
+}
```
aaron-steinfeld, line 30: Up to you, but we could trivially allow writing multiple profiles here. Should fix the Config -> Profile bit in name either way though.

```suggestion
message PutPartitionerProfilesRequest {
  repeated PartitionerProfile profiles = 1;
}
```

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
```diff
@@ -0,0 +1,60 @@
...
+message GetAllPartitionerConfigRequest {
+}
+
+message GetAllPartitionerConfigResponse {
+  repeated PartitionerProfile profiles = 1;
+}
+
+message DeletePartitionerConfigRequest {
+  string profile = 1;
+}
```
aaron-steinfeld, line 44: Similarly, fix name and if you want, can make it plural.

```suggestion
message DeletePartitionerProfilesRequest {
  repeated string profile_names = 1;
}
```

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceImpl.java
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  public void getPartitionerProfile(
+      GetPartitionerProfileRequest request,
+      StreamObserver<GetPartitionerProfileResponse> responseObserver) {
+    try {
+      validator.validateOrThrow(request);
+      Optional<PartitionerProfile> profile =
+          partitionerProfilesStore.getPartitionerProfile(request.getProfile());
+      if (profile.isPresent()) {
```
aaron-steinfeld, line 33: should throw Status.NOT_FOUND (or whatever it's called) if not present, since this is effectively a get by id.

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceImpl.java
```diff
@@ -0,0 +1,86 @@
...
+      } else {
+        responseObserver.onNext(GetPartitionerProfileResponse.newBuilder().build());
+      }
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      responseObserver.onError(e);
+    }
+  }
+
+  /** */
```
aaron-steinfeld, line 45: remove or fill in javadocs

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceImpl.java
```diff
@@ -0,0 +1,86 @@
...
+          partitionerProfilesStore.getPartitionerProfile(request.getProfile());
+      if (profile.isPresent()) {
+        responseObserver.onNext(
+            GetPartitionerProfileResponse.newBuilder().setProfile(profile.get()).build());
+      } else {
+        responseObserver.onNext(GetPartitionerProfileResponse.newBuilder().build());
+      }
+      responseObserver.onCompleted();
+    } catch (Exception e) {
+      responseObserver.onError(e);
```
aaron-steinfeld, line 41: log.warn before returning error.

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,75 @@
...
+      throw Status.INVALID_ARGUMENT
+          .withDescription("partition key cannot be empty")
+          .asRuntimeException();
+    }
+    profile
+        .getGroupsList()
+        .forEach(
+            partitionerGroup -> {
+              if (partitionerGroup.getName().isBlank()) {
+                throw Status.INVALID_ARGUMENT
```
aaron-steinfeld, line 36: this one was informational only right? So it _could_ be. If you don't want it to though, that's fine.

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/store/PartitionerProfileDocument.java
```diff
@@ -0,0 +1,36 @@
...
+    this.partitionerProfile = partitionerProfile;
+  }
+
+  public static PartitionerProfile fromJson(String json) {
+    PartitionerProfile.Builder builder = PartitionerProfile.newBuilder();
+    try {
+      JSON_PARSER.merge(json, builder);
+      return builder.build();
+    } catch (InvalidProtocolBufferException e) {
+      throw new RuntimeException(e);
```
aaron-steinfeld, line 24: Either throw status exceptions here, or catch them later and convert to status.

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/store/PartitionerProfilesDocumentStore.java
```diff
@@ -0,0 +1,59 @@
...
+  @Override
+  public Optional<PartitionerProfile> getPartitionerProfile(String profile) throws Exception {
+    Query query = new Query();
+    query.setFilter(new Filter(Filter.Op.EQ, PARTITIONER_PROFILE_NAME_FIELD, profile));
+    try (CloseableIterator<Document> it = collection.search(query)) {
+      List<Document> list = ImmutableList.copyOf(it);
+      if (list.size() == 1) {
+        return Optional.of(PartitionerProfileDocument.fromJson(list.get(0).toJson()));
+      } else if (list.size() > 1) {
+        throw new IllegalStateException("More than 1 PartitionerProfile documents returned");
```
aaron-steinfeld, line 29: likewise here - always want it to be status before returning, so either throw it directly as a status, or catch and recover/convert later.

aaron-steinfeld (commented):
Mostly api naming nits now that we're not doing it live, but all small things. Mostly lgtm.

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
surajpuvvada: done

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
surajpuvvada: done

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
surajpuvvada: done

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceImpl.java
surajpuvvada: fixed

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceImpl.java
surajpuvvada: done

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceImpl.java
surajpuvvada: removed

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
surajpuvvada: keeping it for now

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/store/PartitionerProfileDocument.java
surajpuvvada: fixed

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/store/PartitionerProfilesDocumentStore.java
surajpuvvada: fixed

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
```diff
@@ -0,0 +1,60 @@
...
+
+message GetPartitionerProfilesRequest {
+}
+
+message GetPartitionerProfilesResponse {
+  repeated PartitionerProfile profiles = 1;
+}
+
+message DeletePartitionerProfilesRequest {
+  repeated string profile = 1;
```
aaron-steinfeld, line 43:
```suggestion
  repeated string profile_names = 1;
```

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,90 @@
...
+
+  public void validateOrThrow(GetPartitionerProfileRequest request) {
+    if (request.getProfileName().isBlank()) {
+      throw Status.INVALID_ARGUMENT.withDescription("profile cannot be empty").asRuntimeException();
+    }
+  }
+
+  public void validateOrThrow(PutPartitionerProfilesRequest request) {
+    request
+        .getProfilesList()
```
aaron-steinfeld, line 18: probably should validate at least one put too

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,90 @@
...
+                    .withDescription("partition_groups cannot be empty")
+                    .asRuntimeException();
+              }
+              if (profile.getPartitionKey().isBlank()) {
+                throw Status.INVALID_ARGUMENT
+                    .withDescription("partition key cannot be empty")
+                    .asRuntimeException();
+              }
+              profile
+                  .getGroupsList()
```
aaron-steinfeld, line 37: at least one group

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,90 @@
...
+                              .asRuntimeException();
+                        }
+                        if (partitionerGroup.getMemberIdsCount() == 0) {
+                          throw Status.INVALID_ARGUMENT
+                              .withDescription("partition_group memberIds cannot be empty")
+                              .asRuntimeException();
+                        }
+                      });
+
+              Map<String, String> memberId2PartitionerGroupName = new HashMap<>();
```
aaron-steinfeld, line 52: nit - avoid using numbers as words

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
```diff
@@ -0,0 +1,90 @@
...
+                                        memberId, partitionerGroup.getName());
+                                  }
+                                });
+                      });
+            });
+  }
+
+  public void validateOrThrow(DeletePartitionerProfilesRequest request) {
+    request
+        .getProfileList()
```
aaron-steinfeld, line 80: here too - more likely to have an invalid request with no values than empty values, so should validate count

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/store/PartitionerProfilesDocumentStore.java
```diff
@@ -0,0 +1,65 @@
...
+          .map(document -> PartitionerProfileDocument.fromJson(document.toJson()))
+          .collect(Collectors.toList());
+    }
+  }
+
+  @Override
+  public void putPartitionerProfiles(List<PartitionerProfile> partitionerProfiles)
+      throws Exception {
+    Map<Key, Document> map = new LinkedHashMap<>();
+    partitionerProfiles.forEach(
```
aaron-steinfeld, line 52: up to you on preference, but since you're using streams for the other transformations, equivalent here would be something like below. I didn't use an ordered map because I didn't see any need for it, but you could switch up the collector to a different ordered one too (like `ImmutableMap.toImmutableMap`)
```java
partitionerProfiles.stream()
  .collect(Collectors.toUnmodifiableMap(PartitionerProfile:: getName, PartitionerProfileDocument::new));
```

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
surajpuvvada: done

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
surajpuvvada: already checking for groupsCount == 0 in L30 above

partitioner-config-service-api/src/main/proto/org/hypertrace/partitioner/config/service/v1/partitioner_config_service.proto
surajpuvvada: done

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/PartitionerConfigServiceRequestValidator.java
surajpuvvada: done

partitioner-config-service-impl/src/main/java/org/hypertrace/partitioner/config/service/store/PartitionerProfilesDocumentStore.java
surajpuvvada: done

aaron-steinfeld (approved):

surajpuvvada:
@aaron-steinfeld how do we override the validate-protos check ? I don't seem to have access to settings

PR #156, opened 2023-01-25
aws s3 bucket channel

japoorv:
## Description
Adding support for AWS S3 bucket channel in notification


notification-channel-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -13,6 +13,13 @@ message NotificationChannelMutableData {
   string channel_name = 1;
   repeated EmailChannelConfig email_channel_config = 2;
   repeated WebhookChannelConfig webhook_channel_config = 3;
+  AwsS3BucketChannelConfig bucket_channel_config = 4;
+}
+
+message AwsS3BucketChannelConfig {
+  string role_arn = 1;
```
japoorv, line 20: `role_arn` is the arn of the role that would be assumed to push the data to bucket (since we have a secret free S3 integration via `WebIdentityToken`)

notification-channel-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -13,6 +13,13 @@ message NotificationChannelMutableData {
   string channel_name = 1;
   repeated EmailChannelConfig email_channel_config = 2;
   repeated WebhookChannelConfig webhook_channel_config = 3;
+  AwsS3BucketChannelConfig bucket_channel_config = 4;
```
aaron-steinfeld, line 16: other destinations are repeated, shouldn't this one be too?

aaron-steinfeld: >  because the way I am thinking about a channel is that it is owned by a team (so Ideally a team should have a single S3 source)

This seems like a logical leap - a channel is designed to aggregate destinations that are interested in a particular notification. That's not necessarily team specific, nor does a team necessarily only want the data in one s3 bucket.

I think the repeated was more for flexibility given the uncertainty. If you model a single bucket with a repeated here, we can always use that as singular up the stack (such as in the UI, for example), and then we have the option later to relax it without breaking anything. If you model a bucket with a singular here, you're forced into breaking the API or carrying tech debt to support multiple later.

japoorv: Okay, makes sense.

notification-channel-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
aaron-steinfeld: Is this the only way we'll ever support authenticating to a bucket? If not, you may want to wrap the details in a one-of so we can support other mechanisms like a basic secret in the future.

notification-channel-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -13,6 +13,13 @@ message NotificationChannelMutableData {
   string channel_name = 1;
   repeated EmailChannelConfig email_channel_config = 2;
   repeated WebhookChannelConfig webhook_channel_config = 3;
+  repeated  AwsS3BucketChannelConfig bucket_channel_config = 4;
```
aaron-steinfeld, line 16:
```suggestion
  repeated AwsS3BucketChannelConfig bucket_channel_config = 4;
```

japoorv:
> May be not necessary, but probably need to update [file1](https://github.com/hypertrace/config-service/blob/main/notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceRequestValidator.java) and take a look at [file2](https://github.com/hypertrace/config-service/blob/main/notification-channel-config-service-impl/src/test/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceImplTest.java)

No, changing the proto files doesn't cause any issues in the prior implementations because it is supposed to be backward compatible. 
I see this issue being mentioned in one another `config-service` PR, in which case we had to manually override and merge the PR (in that case buf was breaking over a new proto module being added, a bug in the plugin), will let @aaron-steinfeld take a look at this

kaushal02:
> > May be not necessary, but probably need to update [file1](https://github.com/hypertrace/config-service/blob/main/notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceRequestValidator.java) and take a look at [file2](https://github.com/hypertrace/config-service/blob/main/notification-channel-config-service-impl/src/test/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceImplTest.java)
> 
> No, changing the proto files doesn't cause any issues in the prior implementations because it is supposed to be backward compatible. 

I was referring to the method [validateNotificationChannelMutableData](https://github.com/hypertrace/config-service/blob/f12463fb0a1329df27464acd0a608b8a81ae33c3/notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceRequestValidator.java#L30) since it checks for the other two members where your proto has changed. But I don't have enough context on the PR, so I'll let you decide whether it's necessary or not.

aaron-steinfeld:
> > May be not necessary, but probably need to update [file1](https://github.com/hypertrace/config-service/blob/main/notification-channel-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceRequestValidator.java) and take a look at [file2](https://github.com/hypertrace/config-service/blob/main/notification-channel-config-service-impl/src/test/java/org/hypertrace/notification/config/service/NotificationChannelConfigServiceImplTest.java)
> 
> No, changing the proto files doesn't cause any issues in the prior implementations because it is supposed to be backward compatible. I see this issue being mentioned in one another `config-service` PR, in which case we had to manually override and merge the PR (in that case buf was breaking over a new proto module being added, a bug in the plugin), will let @aaron-steinfeld take a look at this

So of course you'll want to add validation - but that goes with your impl rather than the api changes. Typically you want the impl and api to go in at the same time though, just reviewed separately.


The actual failure in validate-protos is because your branch is not up to date with main which you've sinced fixed. It was comparing your branch to main and saying "hey, this branch is incompatible with main, it's missing some of the protos that have already been published"

notification-channel-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -13,8 +13,22 @@ message NotificationChannelMutableData {
   string channel_name = 1;
   repeated EmailChannelConfig email_channel_config = 2;
   repeated WebhookChannelConfig webhook_channel_config = 3;
+  repeated AwsS3BucketChannelConfig bucket_channel_config = 4;
```
aaron-steinfeld, line 16: it took me 3 or 4 reads (Sorry!) but I just noticed the name is generic ("bucket"), but the type is s3-specific. If we wanted to add support for GCS next, what would that look like? I think another top level field, so maybe put that specific into this field name too.

japoorv: Sure, adding

notification-channel-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
```diff
@@ -13,8 +13,22 @@ message NotificationChannelMutableData {
   string channel_name = 1;
   repeated EmailChannelConfig email_channel_config = 2;
   repeated WebhookChannelConfig webhook_channel_config = 3;
+  repeated AwsS3BucketChannelConfig s3_bucket_channel_config = 4;
```
aaron-steinfeld, line 16: the proto looks good to me, let's get the validation in and then we can merge.

surajpuvvada (dismissed):
lgtm

notification-channel-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_channel.proto
japoorv: >I'd also ask - any reason not to? The validation code we're discussing is probably a few lines, no more than 10. It'd take less time than it took me to write this comment.

Right, no particular reason of not doing this in this PR. Just wanted to know the logic behind this.

aaron-steinfeld (approved):

PR #157, opened 2023-01-25
chore: add mock deleteAll

SrikarMannepalli:
## Description
This PR adds mock delete all method.

### Checklist:

### DIFF
--- a/object-store/src/main/java/org/hypertrace/config/objectstore/DeletedContextualConfigObjectImpl.java
+++ b/object-store/src/main/java/org/hypertrace/config/objectstore/DeletedContextualConfigObjectImpl.java
@@ -7,13 +7,13 @@ import java.util.function.Function;
 import org.hypertrace.config.service.v1.ContextSpecificConfig;
 
 @lombok.Value
-public class DeletedContextualConfigObjectImpl<T> implements DeletedContextualConfigObject<T> {
+class DeletedContextualConfigObjectImpl<T> implements DeletedContextualConfigObject<T> {
   String context;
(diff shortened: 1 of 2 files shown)


object-store/src/main/java/org/hypertrace/config/objectstore/DeletedContextualConfigObjectImpl.java
```diff
@@ -7,13 +7,13 @@
...
 @lombok.Value
-class DeletedContextualConfigObjectImpl<T> implements DeletedContextualConfigObject<T> {
+public class DeletedContextualConfigObjectImpl<T> implements DeletedContextualConfigObject<T> {
   String context;
   Optional<T> deletedData;
   Instant creationTimestamp;
   Instant lastUpdatedTimestamp;
 
-  DeletedContextualConfigObjectImpl(
+  public DeletedContextualConfigObjectImpl(
```
SrikarMannepalli, line 16: Had to make this public as having to build the object explicitly in some scenarios where delete object method has been overridden.

aaron-steinfeld: Not following - shouldn't be public. If there's cases where we're trying to override this (like in mocks?) they can implement (or mock) the interface.

SrikarMannepalli: Got it. Updated now.

aaron-steinfeld (approved):

PR #161, opened 2023-03-08
Handle validator for Message type descriptor

suresh-prakash:


validation-utils/src/main/java/org/hypertrace/config/validation/GrpcValidatorUtils.java
```diff
@@ -25,8 +26,21 @@ public static <T extends Message> void validateNonDefaultPresenceOrThrow(
 
     if (descriptor.isRepeated()) {
       validateNonDefaultPresenceRepeatedOrThrow(source, descriptor);
-    } else if (!source.hasField(descriptor)
-        || source.getField(descriptor).equals(descriptor.getDefaultValue())) {
+    } else if (descriptor.getType() == Type.MESSAGE
+        && descriptor
+            .getMessageType()
+            .toProto()
+            .getDefaultInstanceForType()
```
aaron-steinfeld, line 33: So why haven't we run into this issue before? I would assume because we typically validate on the leaf values, so sing this on a message wouldn't make a ton of sense. Curious on the use case.

Anyway, I don't think this would work - doesn't this return the default instance for the descriptor proto?

suresh-prakash: Required here: https://github.com/Traceableai/config-service/pull/1060/files#diff-a4c1f2ee59003fb59901991e419e8a42a13d3d7a683b537fba4efa62f2b1eef9

> doesn't this return the default instance for the descriptor proto

Yes. That's what is required and it works as per the tests here: https://github.com/Traceableai/config-service/pull/1060/files#diff-c2366d72b65495cb335bfc3b6476bb795dd091db9277dccb51374883015dc114

In any case, I think, validating just the leaf would work.

PR #163, opened 2023-05-23
Exposing a method to set request context in the mock config service

varsha-abhinandan:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/config-service-api/src/testFixtures/java/org/hypertrace/config/service/test/MockGenericConfigService.java
+++ b/config-service-api/src/testFixtures/java/org/hypertrace/config/service/test/MockGenericConfigService.java
@@ -58,9 +58,9 @@ public class MockGenericConfigService {
 
   private Server grpcServer;
   private Clock clock = Clock.systemUTC();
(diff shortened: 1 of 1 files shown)


config-service-api/src/testFixtures/java/org/hypertrace/config/service/test/MockGenericConfigService.java
```diff
@@ -114,6 +114,11 @@ public MockGenericConfigService withClock(Clock clock) {
     return this;
   }
 
+  public MockGenericConfigService withRequestContext(RequestContext context) {
```
aaron-steinfeld, line 117: So this was really a placeholder context to avoid having to set a context in a test client, but if you actually do care about a context, it would be better to send it the regular way (via client) rather than assigning it in the server. Could we explore that option instead (that might look like changing the test interceptor in this file to only set the context if unset, for example).

aaron-steinfeld (approved):

PR #169, opened 2023-06-13
Added new operator to handle IP not exists in CIDR case

sanket-mundra:
## Description

### Testing

### Checklist:

### Documentation


label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -51,7 +51,8 @@ message LabelApplicationRuleData {
       OPERATOR_UNSPECIFIED = 0;
       OPERATOR_EQUALS = 1; // operator to check if the key exists, and value is equal to the provided value
       OPERATOR_MATCHES_REGEX = 2; // operator to check if the key exists, and value matches the provided regex value
-      OPERATOR_CIDR = 3; // operator to check if the key exists, and IP value lies in the provided CIDR
+      OPERATOR_MATCHES_CIDR = 3; // operator to check if the key exists, and IP value lies in the provided CIDR
```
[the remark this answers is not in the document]
skjindal93: How do we expect a user to specify a list of IPs? CSVs?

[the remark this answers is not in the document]
aaron-steinfeld: > How do we expect a user to specify a list of IPs? CSVs?

UX question. I'm not a huge fan of that, but I think that's how we typically do it. At the API layer and below it should be normalized to either a `String | List<String>` imo.

> How do we expect a user to specify a list of IPs? CSVs?
How about combination of IPs and CIDRs?

My vote would be to support it if it's not prohibitive. Basically something like 
```
IpSpec = <cidr> | <ip>;
Input = IpSpec | List<IpSpec>;
```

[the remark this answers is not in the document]
skjindal93: mocks cleared it out https://www.figma.com/file/l8Q6JjpDfkeA0w6LvQyDWx/API-Catalog?type=design&node-id=4036-28285&t=IMBEwWjqNTKibay2-0

aaron-steinfeld (approved):

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -25,6 +31,8 @@
 import org.hypertrace.label.application.rule.config.service.v1.UpdateLabelApplicationRuleRequest;
 
 public class LabelApplicationRuleValidatorImpl implements LabelApplicationRuleValidator {
+  private static final String STRING_SEPARATOR = ",";
```
aaron-steinfeld, line 34: label rule should not be parsing a comma separated list - we had discussed accepting either a string or list of strings.

ip-utils/src/main/java/org/hypertrace/config/utils/ip/IpValidationUtils.java
```diff
@@ -0,0 +1,163 @@
...
+import org.apache.commons.validator.routines.RegexValidator;
+
+/**
+ * Using 3rd party library <a
+ * href="https://github.com/seancfoley/IPAddress">seancfoley/IPAddress</a>. Validates an ipv6 based
+ * on formats specified over <a
+ * href="https://seancfoley.github.io/IPAddress/IPAddress/apidocs/inet/ipaddr/IPAddressString.html">here</a>.
+ */
+public class IpValidationUtils {
+  private static final String IPV4_REGEX = "^(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})\\.(\\d{1,3})$";
```
aaron-steinfeld, line 19: Why are we manually validating ipv4 in addition to a third party?

ip-utils/src/main/java/org/hypertrace/config/utils/ip/IpValidationUtils.java
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Reason for converting to prefix block explained over <a
+   * href="https://github.com/seancfoley/IPAddress/wiki/Code-Examples-2:-Subnet-Containment,-Matching,-Comparing#containment-within-the-enclosing-cidr-prefix-block">here</a>
+   */
+  public static boolean isValidSubnet(String subnet) {
+    IPAddress parsedAddress = parseAddress(subnet);
+    if (parsedAddress == null) {
+      return false;
+    }
+    return !parsedAddress
```
aaron-steinfeld, line 44: I'm not sure what you're trying to do here, if it parses we should allow it - we don't care if it represents one or multiple. Is this doing something else? If so please include a more descriptive comment rather than a link.

ip-utils/src/main/java/org/hypertrace/config/utils/ip/IpValidationUtils.java
```diff
@@ -0,0 +1,163 @@
...
+  public static boolean isValidIpAddress(String ipAddress) {
+    IPAddress parsedAddress = parseAddress(ipAddress);
+    if (parsedAddress == null) {
+      return false;
+    }
+    return parsedAddress.toPrefixBlock().getCount().equals(ONE);
+  }
+
+  // Caller is expected to validate the arguments
+  public static boolean isIpAddressInRange(String ipRange, String ipAddress) {
```
aaron-steinfeld, line 59: Not sure what the rest of the file is doing, it's no longer validation utils.

ip-utils/src/main/java/org/hypertrace/config/utils/ip/IpValidationUtils.java
```diff
@@ -0,0 +1,163 @@
...
+          return address;
+        }
+        // take ipv4 address part only in case of valid port number
+        return address.substring(0, colonIdx);
+      }
+    }
+    return address;
+  }
+
+  private static boolean isReservedIpAddress(String ipAddress) {
```
aaron-steinfeld, line 115: this is private and not used.

validation-utils/build.gradle.kts
```diff
@@ -6,4 +6,17 @@ dependencies {
   api(libs.hypertrace.grpcutils.context)
   api(libs.grpc.api)
   implementation(libs.protobuf.javautil)
+  implementation(libs.seancfoley.ipaddress)
+
+  compileOnly(libs.lombok)
```
aaron-steinfeld, line 11: lombok no longer used

validation-utils/src/main/java/org/hypertrace/config/validation/IpValidationUtils.java
```diff
@@ -0,0 +1,43 @@
...
+
+  public static boolean isValidIpAddress(String ipAddress) {
+    IPAddress parsedAddress = parseAddress(ipAddress);
+    if (parsedAddress == null) {
+      return false;
+    }
+    return parsedAddress.toPrefixBlock().getCount().equals(ONE);
+  }
+
+  static IPAddress parseAddress(String address) {
```
aaron-steinfeld, line 40: nit - suggest returning an optional since this can be null

validation-utils/src/main/java/org/hypertrace/config/validation/IpValidationUtils.java
```diff
@@ -0,0 +1,43 @@
...
+          // Allows zero-length IPAddressStrings like ""
+          .allowEmpty(false)
+          .toParams();
+
+  public static boolean isValidSubnet(String subnet) {
+    IPAddress parsedAddress = parseAddress(subnet);
+    if (parsedAddress == null) {
+      return false;
+    }
+    return !parsedAddress
```
aaron-steinfeld, line 26: So this is saying if in CIDR notation, we don't allow representing a single IP? If so, why not? Seems like these prefix checks are superfluous, but I'll admit I'm not exactly following

sanket-mundra: > So this is saying if in CIDR notation, we don't allow representing a single IP? If so, why not?

Yes, it is a valid IP address but not a valid subnet. That's how the implementation is in the existing ip-utils.

aaron-steinfeld: The great thing about not having a shared lib here is we can tailor this to what's relevant here. When the use case allows either, it's silly to parse it twice to figure out if it's valid. It's more compute and it's harder to read the validator code both here and where it's called.

So this whole file could really just be
```java
boolean isValidIpAddressOrSubnet(String input) {
  return new IPAddressString(address, ADDRESS_VALIDATION_PARAMS).getAddress() != null;
}
```

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -120,7 +135,26 @@ private void validateJsonCondition(JsonCondition jsonCondition) {
   }
 
   private void validateStringCondition(StringCondition stringCondition) {
-    validateNonDefaultPresenceOrThrow(stringCondition, stringCondition.OPERATOR_FIELD_NUMBER);
+    validateNonDefaultPresenceOrThrow(stringCondition, StringCondition.OPERATOR_FIELD_NUMBER);
+    if (stringCondition.getOperator().equals(OPERATOR_MATCHES_IPS)
+        || stringCondition.getOperator().equals(OPERATOR_NOT_MATCHES_IPS)) {
+      if (stringCondition.getKindCase().equals(StringCondition.KindCase.VALUE)) {
+        validateNonDefaultPresenceOrThrow(stringCondition, StringCondition.VALUE_FIELD_NUMBER);
```
aaron-steinfeld, line 142: not needed, the case you checked on line 141 checks that this is set.

Might be a bit more readable as a switch in case other cases are added in future. My personal pref is to always use switches when working with enums to avoid that assumption that it will never be added to (and then static analysis can flag missing cases). Here for example, imagine neither value nor values was set. You'd go into the else case assuming it must be values, and then get a cryptic message that the plural version is not set (rather than that the string condition has neither set, which the default case of the switch would throw).

Anyway, not required to do the switch just something to think about.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule.proto
```diff
@@ -45,13 +45,21 @@ message LabelApplicationRuleData {
 
   message StringCondition {
     Operator operator = 1;
-    string value = 2;
+    oneof kind {
+      string value = 2;
```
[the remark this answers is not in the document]
sanket-mundra: Verified both the scenarios!

aaron-steinfeld:
Overriding proto validation check since we've verified this is a compatible change.

PR #178, opened 2023-10-11
chore | delete configs permanently rather than soft deletes

aman-bansal:


config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -166,6 +174,20 @@ public ContextSpecificConfig getConfig(ConfigResourceContext configResourceConte
...
+      throws IOException {
+    Map<ConfigResourceContext, Optional<ConfigDocument>> configDocs =
+        getLatestVersionConfigDocs(configResourceContexts);
+    return configDocs.values().stream()
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .map(this::convertToContextSpecificConfig)
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .collect(Collectors.toUnmodifiableList());
```
SrikarMannepalli, line 188:
```suggestion
    return configDocs.values().stream()
        .flatMap(Optional::stream)
        .map(this::convertToContextSpecificConfig)
        .flatMap(Optional::stream)
        .collect(Collectors.toUnmodifiableList());
```

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -158,6 +158,14 @@ public List<UpsertedConfig> writeAllConfigs(
     return this.writeConfigs(resourceContextValueMap, userId);
   }
 
+  @Override
+  public void deleteConfigs(Set<ConfigResourceContext> configResourceContexts) {
+    if (configResourceContexts.isEmpty()) {
+      return;
+    }
+    collection.delete(getConfigResourceContextsFilter(configResourceContexts));
```
SrikarMannepalli, line 166: Should we also return and maybe log the total deleted documents returned by this doc store method? 🤔

gradle/libs.versions.toml
```diff
@@ -2,9 +2,9 @@
 protoc = "3.23.2"
 grpc = "1.57.2"
 gson = "2.9.0"
-hypertrace-grpcutils = "0.12.2"
-hypertrace-framework = "0.1.58"
-kafka-streams-framework = "0.3.2"
+hypertrace-grpcutils = "0.12.5"
+hypertrace-framework = "0.1.61"
+kafka-streams-framework = "0.3.9"
```
SrikarMannepalli, line 7: Let's also bump up document store.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
aman-bansal: we are returning the documents before after get. Also delete doesnt return the count. it only returns the boolean.

SrikarMannepalli: Ah, got it. I think bulk delete returns `BulkDeleteResult` which has the deleted count.

SrikarMannepalli (approved):

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
```diff
@@ -42,6 +43,15 @@ UpsertedConfig writeConfig(
    */
   ContextSpecificConfig getConfig(ConfigResourceContext configResourceContext) throws IOException;
 
+  /**
+   * Get the configs with the latest version for the specified resources.
+   *
+   * @param configResourceContexts
+   * @return
+   */
+  List<ContextSpecificConfig> getConfigs(Set<ConfigResourceContext> configResourceContexts)
```
skjindal93, line 52:
```suggestion
  List<ContextSpecificConfig> getAllConfigs(Set<ConfigResourceContext> configResourceContexts)
```

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
```diff
@@ -42,6 +43,15 @@ UpsertedConfig writeConfig(
    */
   ContextSpecificConfig getConfig(ConfigResourceContext configResourceContext) throws IOException;
 
+  /**
+   * Get the configs with the latest version for the specified resources.
```
skjindal93, line 47: Can we have a similar Java doc comment as L56-57, if that holds true?

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -254,6 +262,22 @@ private Map<ConfigResourceContext, Optional<ConfigDocument>> getLatestVersionCon
     return latestVersionConfigDocs;
   }
 
+  private Filter getConfigResourceContextsFilter(
```
skjindal93, line 265: nit:
```suggestion
  private Filter buildConfigResourceContextsFilter(
```

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -155,39 +154,23 @@ public void deleteConfigs(
...
+      // delete the configs for the specified config resources.
+      configStore.deleteConfigs(configResourceContexts);
       responseObserver.onNext(
-          DeleteConfigsResponse.newBuilder()
-              .addAllDeletedConfigs(
-                  deletedConfigs.stream()
-                      .map(this::buildDeletedContextSpecificConfig)
-                      .collect(Collectors.toUnmodifiableList()))
-              .build());
+          DeleteConfigsResponse.newBuilder().addAllDeletedConfigs(configs).build());
```
skjindal93, line 166: Do we need to return the entire set of deleted configs? Or, just the config ids?

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -155,39 +154,23 @@ public void deleteConfigs(
...
-              .collect(ImmutableMap.toImmutableMap(Entry::getKey, Entry::getValue));
-
-      List<UpsertedConfig> deletedConfigs =
-          configStore.writeAllConfigs(valuesByContext, getUserId());
 
+      Set<ConfigResourceContext> configResourceContexts =
+          request.getConfigsList().stream()
+              .map(this::getConfigResourceContext)
+              .collect(Collectors.toUnmodifiableSet());
+      List<ContextSpecificConfig> configs = configStore.getConfigs(configResourceContexts);
```
skjindal93, line 162: Is it possible to re-use `getAllConfigs#ResourceConfig`, instead of exposing the new function?

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aman-bansal: contract says the complete config. thats why made the change to not break the contract

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
skjindal93: got it. I was confused about the implementation of `getAllConfigs`. We can think about calling `getAllContextConfigs`, but the current naming is fine too

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
skjindal93: I was actually confused between `getAllConfigs` and `getConfigs`. They looked the same based on the naming.

I understand the differentiation now

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
skjindal93, line 52: What would be the order of the configs returned? 
- Should we consider taking in a list as an input?
OR
- Should we consider returning map as an output?

skjindal93 (dismissed):

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -125,16 +125,15 @@ public void deleteConfig(
     try {
       ConfigResourceContext configResourceContext = getConfigResourceContext(request);
       ContextSpecificConfig configToDelete = configStore.getConfig(configResourceContext);
-
       // if configToDelete is null/empty (i.e. config value doesn't exist or is already deleted),
       // then throw NOT_FOUND exception
-      if (ConfigServiceUtils.isNull(configToDelete.getConfig())) {
+      if (configToDelete == null || ConfigServiceUtils.isNull(configToDelete.getConfig())) {
```
aaron-steinfeld, line 130: configStore.getConfig has historically been a non-nullable response and it doesn't look like that's changed. If you want to change it, make it return an optional instead.

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
```diff
@@ -42,6 +43,17 @@ UpsertedConfig writeConfig(
...
 
+  /**
+   * Get all the configs with the latest version(along with the context to which it applies) for the
+   * specified resource keys
+   *
+   * @param configResourceContexts
+   * @return the configs
+   * @throws IOException
+   */
+  Map<ConfigResourceContext, ContextSpecificConfig> getAllContextConfigs(
```
aaron-steinfeld, line 54:
```suggestion
  Map<ConfigResourceContext, ContextSpecificConfig> getContextConfigs(
```

It's a plural get, but you're filtering so it's not a "get all".

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
```diff
@@ -42,6 +43,17 @@ UpsertedConfig writeConfig(
...
+      Set<ConfigResourceContext> configResourceContexts) throws IOException;
```
aaron-steinfeld, line 55: Accept liberally. No reason that this needs to be a set, make it a collection (same on plural delete below).

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -254,6 +268,22 @@ private Map<ConfigResourceContext, Optional<ConfigDocument>> getLatestVersionCon
...
+        configResourceContexts.stream()
+            .map(this::getConfigResourceFieldContextFilter)
+            .collect(Collectors.toUnmodifiableList());
+    Filter configResourceFieldContextFilter = new Filter();
+    configResourceFieldContextFilter.setOp(Filter.Op.OR);
+    configResourceFieldContextFilter.setChildFilters(childFilters.toArray(Filter[]::new));
+    Filter tenantIdFilter =
+        Filter.eq(
+            TENANT_ID_FIELD_NAME,
+            configResourceContexts.iterator().next().getConfigResource().getTenantId());
```
aaron-steinfeld, line 283: Confusing read. suggest `Iterables.getFirst`

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -254,6 +268,22 @@ private Map<ConfigResourceContext, Optional<ConfigDocument>> getLatestVersionCon
...
   }
 
+  private Filter buildConfigResourceContextsFilter(
+      Set<ConfigResourceContext> configResourceContexts) {
+    List<Filter> childFilters =
+        configResourceContexts.stream()
+            .map(this::getConfigResourceFieldContextFilter)
+            .collect(Collectors.toUnmodifiableList());
+    Filter configResourceFieldContextFilter = new Filter();
+    configResourceFieldContextFilter.setOp(Filter.Op.OR);
```
aaron-steinfeld, line 278: Note this will be incorrect for an empty set - it would build an "all" filter rather than a "none". You're checking outside, which is good, but you should check in here too so future usages can't make that mistake.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -166,6 +175,23 @@ public ContextSpecificConfig getConfig(ConfigResourceContext configResourceConte
         .orElseGet(() -> emptyConfig(configResourceContext.getContext()));
   }
 
+  @Override
+  public Map<ConfigResourceContext, ContextSpecificConfig> getAllContextConfigs(
+      Set<ConfigResourceContext> configResourceContexts) throws IOException {
+    Map<ConfigResourceContext, Optional<ConfigDocument>> configDocs =
+        getLatestVersionConfigDocs(configResourceContexts);
+
+    return configDocs.entrySet().stream()
```
aaron-steinfeld, line 184: Consider Maps.transformValues

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -166,6 +175,23 @@ public ContextSpecificConfig getConfig(ConfigResourceContext configResourceConte
...
+        .collect(
+            Collectors.toUnmodifiableMap(
+                Entry::getKey,
+                entry ->
+                    entry
+                        .getValue()
+                        .flatMap(this::convertToContextSpecificConfig)
+                        .orElseGet(() -> emptyConfig(entry.getKey().getContext()))));
```
aaron-steinfeld, line 192: Do you need to do this? Could you just return a map of the contexts that actually exist?

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
aman-bansal: problem would be null values. i need to filter it out later

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
aman-bansal: can add filter let me try you can review then

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
aaron-steinfeld: If the response has a different keyset than the source, then yes you'd probably use a collector. In the current code at least, that's not the case - you're strictly transforming the value.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
aaron-steinfeld: If you're trying to build a filter that matches N number of IDs, and N is 0, it should return no results. Here it returns _all_ results in the 0 case. If we had implemented delete without an empty check, and someone called it with an empty list it would drop the whole collection (for that tenant/resource).

```
collection.delete(buildConfigResourceContextsFilter(configResourceContexts));
```

In other words, there's no exposed bug here, but that relies on callers doing validation which can easily be done in here.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -66,6 +66,12 @@ public void init(Config config) {
     this.collection = this.datastore.getCollection(CONFIGURATIONS_COLLECTION);
   }
 
+  @VisibleForTesting
```
aaron-steinfeld, line 69: Pretty much always an anti-pattern. there's no DI in this module right? Suggest removing this "init" method and just have the constructor take in a Collection and a Clock. The construction of a datastore shouldn't be the responsibility of this class.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -69,17 +71,24 @@ public void upsertConfig(
   public void getConfig(
       GetConfigRequest request, StreamObserver<GetConfigResponse> responseObserver) {
     try {
-      ContextSpecificConfig config = configStore.getConfig(getConfigResourceContext(request));
-
+      ConfigResourceContext configResourceContext = getConfigResourceContext(request);
+      Optional<ContextSpecificConfig> maybeConfig = configStore.getConfig(configResourceContext);
+      // initialize empty config if not present to handle merges better
+      if (maybeConfig.isEmpty()) {
+        maybeConfig = Optional.of(emptyConfig(configResourceContext.getContext()));
```
aaron-steinfeld, line 78: At this point, you know config is present. Don't separate presence checks and access like this, instead you could for example write:
```java

ContextSpecificConfig config = configStore.getConfig(configResourceContext)
  .orElse(emptyConfig(configResourceContext.getContext()));
```

Then all of your later accesses can be type safe.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -69,17 +71,24 @@ public void upsertConfig(
...
+      }
       // get the configs for the contexts mentioned in the request and merge them in the specified
       // order
       for (String context : request.getContextsList()) {
-        ContextSpecificConfig contextSpecificConfig =
+        Optional<ContextSpecificConfig> contextSpecificConfig =
             configStore.getConfig(getConfigResourceContext(request, context));
-        config = merge(config, contextSpecificConfig);
+        if (contextSpecificConfig.isEmpty()) {
```
aaron-steinfeld, line 85: Same thing here. If you find yourself using .get(), .isEmpty() etc. you're likely misusing an optional.
```java
config = maybeContextConfig
  .map(contextConfig -> merge(config, contextConfig))
  .orElse(config);
```

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -124,18 +133,18 @@ public void deleteConfig(
       DeleteConfigRequest request, StreamObserver<DeleteConfigResponse> responseObserver) {
     try {
       ConfigResourceContext configResourceContext = getConfigResourceContext(request);
-      ContextSpecificConfig configToDelete = configStore.getConfig(configResourceContext);
+      Optional<ContextSpecificConfig> maybeConfig = configStore.getConfig(configResourceContext);
```
aaron-steinfeld, line 136: suggest mapping this one to a response object. Then, if response is present, delete configs and return; else error.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -155,16 +164,21 @@ public void deleteConfigs(
...
-      Set<ConfigResourceContext> configResourceContexts =
+      List<ConfigResourceContext> configResourceContexts =
           request.getConfigsList().stream()
               .map(this::getConfigResourceContext)
-              .collect(Collectors.toUnmodifiableSet());
-      Map<ConfigResourceContext, ContextSpecificConfig> configs =
-          configStore.getAllContextConfigs(configResourceContexts);
+              .collect(Collectors.toUnmodifiableList());
+      Map<ConfigResourceContext, Optional<ContextSpecificConfig>> configs =
+          configStore.getContextConfigs(configResourceContexts);
```
aaron-steinfeld, line 172: Do you need to return a Map<context, optional<config>> ? Why not just remove the contexts that aren't present from the map entirely. Then you can delete only the keyset to save a bit of work.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aman-bansal: thanks i will keep this in mind while using optional

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aman-bansal: We returned this because there is dependency of empty config in multiple places across config service. so Optional<ContextSpecificConfig> will sometimes have non empty value but config inside is empty. I didn't filter it out because it might lead to missing some configs which are null and present in DB

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aman-bansal: i didnt get what you mean by this?

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aaron-steinfeld: true, but can easily copy the ref like
```
ContextSpecificConfig lastConfig = config;
config = maybeContextConfig
  .map(contextConfig -> merge(lastConfig, contextConfig))
  .orElse(config);
```

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aaron-steinfeld: So I don't think we should mix optional and empty config (if I'm understanding you correctly here). If we're switching to optionals, we should be replacing the empties. We don't want two ways to represent "no config". It might make sense to back out the change from empty -> optional and do that in a separate PR.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aaron-steinfeld: e.g.
```java
    try {
      ConfigResourceContext configResourceContext = getConfigResourceContext(request);
      DeleteConfigResponse deleteResponse = configStore.getConfig(configResourceContext)
          .map(configToDelete -> DeleteConfigResponse.newBuilder().setDeletedConfig(configToDelete).build())
          .orElseThrow(Status.NOT_FOUND::asException);

      // delete the config for the specified config resource.
      configStore.deleteConfigs(Set.of(configResourceContext));
      responseObserver.onNext(deleteResponse);
    } catch (Exception e) {
      log.error("Delete config failed for request:{}", request, e);
      responseObserver.onError(e);
    }
```

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aman-bansal: yeah this just looks weird to me that's why

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aman-bansal: got it

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
aman-bansal: let me track this via separate JIRA. I will create a separate PR for this. That might be big change in terms of how we handle configs right now in terms of creation and updates

config-service-impl/src/test/java/org/hypertrace/config/service/ConfigServiceGrpcImplTest.java
```diff
@@ -215,30 +218,21 @@ void deleteConfigs() throws IOException {
...
-            DeleteConfigsResponse.newBuilder()
-                .addAllDeletedConfigs(
-                    List.of(
-                        buildContextSpecificConfig(context1, config1, 10L, 20L),
-                        buildContextSpecificConfig(context2, config2, 10L, 20L)))
-                .build());
+    verify(configStore, times(1))
+        .deleteConfigs(
+            eq(Set.of(getConfigResourceContext(context1), getConfigResourceContext(context2))));
+    verify(responseObserver, times(1)).onNext(any());
```
aaron-steinfeld, line 235: Please restore the response verification. This shouldn't change if we haven't broken compat

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -42,36 +41,22 @@
...
+  private final Collection collection;
 
-  public DocumentConfigStore() {
-    this(Clock.systemUTC());
+  public DocumentConfigStore(Datastore datastore) {
+    this(Clock.systemUTC(), datastore);
   }
 
-  DocumentConfigStore(Clock clock) {
+  DocumentConfigStore(Clock clock, Datastore datastore) {
```
aaron-steinfeld, line 54: nit - make this the public one and remove above

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigStore.java
```diff
@@ -40,7 +33,19 @@ UpsertedConfig writeConfig(
...
+
+  /**
+   * Get all the configs with the latest version(along with the context to which it applies) for the
+   * specified resource keys
+   *
+   * @param configResourceContexts
+   * @return the configs
+   * @throws IOException
+   */
+  Map<ConfigResourceContext, Optional<ContextSpecificConfig>> getContextConfigs(
```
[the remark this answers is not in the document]
aaron-steinfeld: old, unsubmitted comment

aaron-steinfeld (approved):

PR #180, opened 2023-10-14
feat: added syslog support in notification channel

singhalprerana:
Added support for syslog server integration in notification channel config.


config-service-impl/src/test/java/org/hypertrace/config/service/ConfigServiceGrpcImplTest.java
```diff
@@ -232,14 +233,7 @@ void deleteConfigs() throws IOException {
...
             eq(Set.of(getConfigResourceContext(context1), getConfigResourceContext(context2))));
-    verify(responseObserver, times(1))
-        .onNext(
-            DeleteConfigsResponse.newBuilder()
-                .addAllDeletedConfigs(
-                    List.of(
-                        buildContextSpecificConfig(context1, config1, 10L, 20L),
-                        buildContextSpecificConfig(context2, config2, 10L, 20L)))
-                .build());
+    verify(responseObserver, times(1)).onNext(argThat(this::matchDeletedResponse));
```
singhalprerana, line 236: this test was flaky - was failing intermittently, so fixed it (though unrelated to this PR)

PR #189, opened 2023-12-14
Validate regex in label rule creation/updation

sanket-mundra:


label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/RegexValidator.java
```diff
@@ -0,0 +1,19 @@
+package org.hypertrace.label.application.rule.config.service;
+
+import com.google.re2j.Pattern;
+import com.google.re2j.PatternSyntaxException;
+import io.grpc.Status;
+
+public class RegexValidator {
```
skjindal93, line 7: `validation-utils` module should be a good candidate for this class

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -123,31 +122,55 @@ private void validateJsonCondition(JsonCondition jsonCondition) {
     }
   }
 
+  private void validateKeyStringCondition(StringCondition stringCondition) {
+    validateNonDefaultPresenceOrThrow(stringCondition, StringCondition.OPERATOR_FIELD_NUMBER);
```
aaron-steinfeld, line 126: you're already validating this by switching on the value of operator next.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -123,31 +122,55 @@ private void validateJsonCondition(JsonCondition jsonCondition) {
     }
   }
 
+  private void validateKeyStringCondition(StringCondition stringCondition) {
```
aaron-steinfeld, line 125: shouldn't this be validating the value in addition to the operator? In general, why does it need a different validator than the general string condition?

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -123,31 +122,55 @@ private void validateJsonCondition(JsonCondition jsonCondition) {
...
-                "StringCondition in LabelApplicationRule should have a valid IP address or CIDR");
-          }
-          break;
-        case VALUES:
-          if (stringCondition.getValues().getValuesList().stream()
-              .anyMatch(s -> !isValidIpAddressOrSubnet(s))) {
+    switch (stringCondition.getOperator()) {
+      case OPERATOR_MATCHES_REGEX:
+        final String pattern = stringCondition.getValue();
+        Status regexValidationStatus = RegexValidator.validate(pattern);
```
aaron-steinfeld, line 144: Does an empty string pass validation here (it shouldn't)?

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -123,31 +122,55 @@ private void validateJsonCondition(JsonCondition jsonCondition) {
...
+        }
+        break;
+      case OPERATOR_MATCHES_IPS:
+      case OPERATOR_NOT_MATCHES_IPS:
+        switch (stringCondition.getKindCase()) {
+          case VALUE:
+            final String ip = stringCondition.getValue();
+            if (!isValidIpAddressOrSubnet(ip)) {
+              throwInvalidArgumentException(
+                  "StringCondition in LabelApplicationRule should have a valid IP address or CIDR");
```
aaron-steinfeld, line 158: nit - would be helpful to include the invalid condition in the exception

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
sanket-mundra: Yes missed validating the value. It needs a different validator because we support only equals and regex match for key condition. Thus, adding separate validations for it. Ideally we should have used a different Operator enum for key condition.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
sanket-mundra: added validations

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
aaron-steinfeld: Got it

>  Ideally we should have used a different Operator enum for key condition.

Probably other way around :) key condition was here before the ip types were recently added.

aaron-steinfeld (approved):

PR #209, opened 2024-04-23
Changes in Notification rule to allow types

DibyojyotiS:
## Description
For non-streaming notification. Particularly for integrations supporting a batched type of notification.

### Testing

### Checklist:

### Documentation

### DIFF
--- a/notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
+++ b/notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
@@ -30,11 +30,6 @@ message NotificationChannelTarget {
 }
 
 message NotificationIntegrationTarget {
-  IntegrationType type = 1;
+  string type = 1;
   string integration_id = 2;
-}
-
-enum IntegrationType {
-  INTEGRATION_TYPE_UNSPECIFIED = 0;
-  INTEGRATION_TYPE_WIZ = 1;
 }
--- a/notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
+++ b/notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
@@ -19,12 +19,22 @@ message NotificationRuleMutableData {
   string channel_id = 5 [deprecated=true];
   string rate_limit_interval_duration = 6;
 
-  string sink_id = 7;
-  SinkType sink_type = 8;
+  oneof notification_target {
+    NotificationChannelTarget channel_target = 7;
+    NotificationIntegrationTarget integration_target = 8;
+  }
 }
 
-enum SinkType {
-  SINK_TYPE_UNSPECIFIED = 0;
-  SINK_TYPE_CHANNEL = 1;
-  SINK_TYPE_WIZ_INTEGRATION = 2;
+message NotificationChannelTarget {
+  string channel_id = 1;
+}
+
+message NotificationIntegrationTarget {
+  IntegrationType type = 1;
+  string integration_id = 2;
+}
+
+enum IntegrationType {
+  INTEGRATION_TYPE_UNSPECIFIED = 0;
+  INTEGRATION_TYPE_WIZ = 1;
 }
--- a/notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
+++ b/notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
@@ -16,6 +16,15 @@ message NotificationRuleMutableData {
   string event_condition_id = 3;
   string event_condition_type = 4;
 
-  string channel_id = 5;
+  string channel_id = 5 [deprecated=true];
   string rate_limit_interval_duration = 6;
+
+  string sink_id = 7;
+  SinkType sink_type = 8;
+}
+
+enum SinkType {
+  SINK_TYPE_UNSPECIFIED = 0;
+  SINK_TYPE_CHANNEL = 1;
+  SINK_TYPE_WIZ_INTEGRATION = 2;
 }
--- a/notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
+++ b/notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
@@ -16,17 +16,10 @@ message NotificationRuleMutableData {
   string event_condition_id = 3;
   string event_condition_type = 4;
 
-  string channel_id = 5 [deprecated=true];
+  string channel_id = 5;
   string rate_limit_interval_duration = 6;
 
-  oneof notification_target {
-    NotificationChannelTarget channel_target = 7;
-    NotificationIntegrationTarget integration_target = 8;
-  }
-}
-
-message NotificationChannelTarget {
-  string channel_id = 1;
+  NotificationIntegrationTarget integration_target = 7;
 }
 
 message NotificationIntegrationTarget {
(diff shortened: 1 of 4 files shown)


japoorv (dismissed):

notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
```diff
@@ -16,6 +16,15 @@ message NotificationRuleMutableData {
   string event_condition_id = 3;
   string event_condition_type = 4;
 
-  string channel_id = 5;
+  string channel_id = 5 [deprecated=true];
```
singhalprerana, line 19: we should have something like
```
oneof notification_target {
    NotificationChannel channel = 1;
    NotificationIntegrationTarget integration = 2;
  }

  message NotificationChannel {
    string channel_id = 1;
  }

  message NotificationIntegrationTarget {
    IntegrationType type = 1;
    string integration_id = 2;
  }

  enum IntegrationType {
    WIZ = 1;
  }
```

DibyojyotiS: @aaron-steinfeld Curious to know your thoughts on this.

singhalprerana:
@aaron-steinfeld the policy-engine changes are still some time away and this is a priority requirement, hence we need to do the changes now.. cc: @avinashkolluru

singhalprerana (approved):

notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
```diff
@@ -16,6 +16,25 @@ message NotificationRuleMutableData {
...
   string rate_limit_interval_duration = 6;
+
+  oneof notification_target {
```
singhalprerana, line 22: @aaron-steinfeld Based on the slack discussion (pasted the snippet below), hope we can go forward with this?

> now one particular notification condition requires a special notification config

that’s how the requirement goes..

> Are they actually agnostic though? every notification type can be sent to every channel already?

yes AFAIK and from the user UI perspective too

> Agree an enum is better here, but since both would be placeholders until we migrate to policy engine, it doesn’t seem like it’s a significant amount of tech debt to take on. The whole idea of a channel is the real tech debt.

if that’s anyways true, there shouldn’t be any harm to have an enum.. that would be clearer of the 2 options..

DibyojyotiS: > if that’s anyways true, there shouldn’t be any harm to have an enum.. that would be clearer of the 2 options..
Which Enum are we referring to?

aaron-steinfeld: > Which Enum are we referring to?

The integration type - my preference was to use a string there so this notification is not tightly coupled with integration support as it's just one more place to change when adding a new type (since the only code that actually needs to know about the type is the code that's evaluating the rule).

singhalprerana: @aaron-steinfeld currently channel allows co-existence of multiple components - say email+webhook+splunk.. But in this case, Wiz cannot coexist with these other components because of its different behaviour. I was preferring keeping this restriction as part of the API rather than the channel implementation..

aaron-steinfeld: So an event being pushed to wiz cannot be sent as an email? If so, I think it's fine to leave it in the rule,  but my preference would be something like

```proto
message NotificationRuleMutableData {
  string rule_name = 1;
  string description = 2;

  string event_condition_id = 3;
  string event_condition_type = 4;

  optional string channel_id = 5;
  string rate_limit_interval_duration = 6;
  NotificationIntegrationTarget integration_target = 7;
}

message NotificationIntegrationTarget {
  string type = 1;
  string integration_id = 2;
}

```

singhalprerana: > You can add the integration target as a sibling and un-deprecate the channel id to dramatically reduce the blast radius of this change.

Here can we leave channel_id as is and introduce channelType, something like:
```
message ChannelType {
  CHANNEL_TYPE_COMBINATION = 0;
  CHANNEL_TYPE_WIZ_INTEGRATION = 1;
}
```
I know it's a bit hacky, but @aaron-steinfeld does this work? then `channel_id` can be reused as integration_id in wiz case..

singhalprerana: ^^^ works for me.. missed the part that we can add optional to string without breaking API.. @DibyojyotiS let's go with this..

aaron-steinfeld: it's _technically_ a breaking change, but I think the functional impact is less than if we left it deprecated and had a second way of referencing a channel.

DibyojyotiS: I made the change - replaced the enum with a string.

aaron-steinfeld: The second part was unwinding the one of and leaving the channel field as is, just adding the optional qualifier - see above

DibyojyotiS: > it's technically a breaking change, but I think the functional impact is less than if we left it deprecated and had a second way of referencing a channel.

Oh, my bad. I had incorrectly read this comment - somehow I missed the word "than" and it changed the meaning. Removed the oneof too.

aaron-steinfeld (dismissed):

notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleConfigServiceRequestValidator.java
```diff
@@ -28,7 +29,14 @@ public void validateUpdateNotificationRuleRequest(
 
   private void validateNotificationRuleMutableData(NotificationRuleMutableData data) {
     validateNonDefaultPresenceOrThrow(data, NotificationRuleMutableData.RULE_NAME_FIELD_NUMBER);
-    validateNonDefaultPresenceOrThrow(data, NotificationRuleMutableData.CHANNEL_ID_FIELD_NUMBER);
+    if (data.hasIntegrationTarget()) {
```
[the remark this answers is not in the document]
DibyojyotiS: done

aaron-steinfeld (approved):

notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule.proto
```diff
@@ -16,6 +16,13 @@ message NotificationRuleMutableData {
...
 
-  string channel_id = 5;
+  optional string channel_id = 5;
   string rate_limit_interval_duration = 6;
+
+  NotificationIntegrationTarget integration_target = 7;
+}
+
+message NotificationIntegrationTarget {
+  string type = 1;
```
singhalprerana, line 26: I am not fully onboard with this - still feel enum would be better for a single source of truth. But given it's a small nuance in the scheme of things and it works for our use-case, so works for me.

singhalprerana (approved):

singhalprerana:
@aaron-steinfeld approved - works for the current use-case. Plz go ahead and merge.

PR #210, opened 2024-05-10
chore: add new validation util and a bit of housekeeping

aaron-steinfeld:
## Description

Add a validation util that detects whether an optional field has been explicitly assigned. Our existing util looks if a field resolves to its default value which means it will flag optional fields assigned to their default value - sometimes that's desirable, but sometimes it's not.

Did a bit of housekeeping with old suppressions and versions while I was in here.

### DIFF
--- a/.github/workflows/pr-build.yml
+++ b/.github/workflows/pr-build.yml
@@ -13,14 +13,14 @@ jobs:
     steps:
       # Set fetch-depth: 0 to fetch commit history and tags for use in version calculation
       - name: Check out code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           ref: ${{github.event.pull_request.head.ref}}
           repository: ${{github.event.pull_request.head.repo.full_name}}
           fetch-depth: 0
 
       - name: Login to Docker Hub
-        uses: docker/login-action@v2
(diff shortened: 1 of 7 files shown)


.github/workflows/pr-build.yml
```diff
@@ -41,14 +41,14 @@ jobs:
...
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
         with:
           ref: ${{github.event.pull_request.head.ref}}
           repository: ${{github.event.pull_request.head.repo.full_name}}
           fetch-depth: 0
 
       - name: Setup buf
-        uses: bufbuild/buf-setup-action@v1.9.0
+        uses: bufbuild/buf-setup-action@v1
```
tim-mwangi, line 51: Is this a downgrade?

aaron-steinfeld: No - I just unpinned the minor/patch version (leaving major version pinned) to keep it consistent with the next two steps. The latest is 1.31.0 so this is effectively an upgrade.

tim-mwangi: I thought so. Just wanted to make sure.

tim-mwangi (approved):

PR #212, opened 2024-05-29
Store user details in config objects in mongo

Vaibhav090420:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -39,6 +39,7 @@ message UpsertConfigRequest {
   // If empty, specified config is associated with a default context.
   string context = 4;
 
+  // user details
   UserDetails user_details = 5;
 }
 
@@ -55,6 +56,7 @@ message UpsertConfigResponse {
   // prev version of config value in the store
   optional google.protobuf.Value prev_config = 4;
 
+  // user details
   UserDetails user_details = 5;
 }
 
@@ -157,6 +159,7 @@ message UpsertAllConfigsRequest {
 message UpsertAllConfigsResponse {
   repeated UpsertedConfig upserted_configs = 1;
   UserDetails user_details = 2;
+
   message  UpsertedConfig {
     string context = 1;
     google.protobuf.Value config = 2;
@@ -167,5 +170,5 @@ message UpsertAllConfigsResponse {
 }
 
 message UserDetails {
-  string userEmail = 1;
+  string user_email = 1;
 }
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -38,6 +38,8 @@ message UpsertConfigRequest {
   // optional - only required if config applies to a specific context.
   // If empty, specified config is associated with a default context.
   string context = 4;
+
+  UserDetails user_details = 5;
 }
 
 message UpsertConfigResponse {
@@ -52,6 +54,8 @@ message UpsertConfigResponse {
 
   // prev version of config value in the store
   optional google.protobuf.Value prev_config = 4;
+
+  UserDetails user_details = 5;
 }
 
 message GetConfigRequest {
@@ -141,7 +145,7 @@ message DeleteConfigsResponse {
 
 message UpsertAllConfigsRequest {
   repeated ConfigToUpsert configs = 1;
-
+  UserDetails user_details = 2;
   message ConfigToUpsert {
     string resource_name = 1;
     string resource_namespace = 2;
@@ -152,12 +156,16 @@ message UpsertAllConfigsRequest {
 
 message UpsertAllConfigsResponse {
   repeated UpsertedConfig upserted_configs = 1;
-
-  message UpsertedConfig {
+  UserDetails user_details = 2;
+  message  UpsertedConfig {
     string context = 1;
     google.protobuf.Value config = 2;
     int64 creation_timestamp = 3;
     int64 update_timestamp = 4;
     optional google.protobuf.Value prev_config = 5;
   }
+}
+
+message UserDetails {
+  string userEmail = 1;
 }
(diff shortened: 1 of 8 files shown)


config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -38,6 +38,9 @@ message UpsertConfigRequest {
   // optional - only required if config applies to a specific context.
   // If empty, specified config is associated with a default context.
   string context = 4;
+
+  // user details
+  UserDetails user_details = 5;
```
aaron-steinfeld, line 43: Who is the user of a config? And assuming you mean the user making this change, this does not belong in the API - it must be coming from the context.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -52,6 +55,9 @@ message UpsertConfigResponse {
 
   // prev version of config value in the store
   optional google.protobuf.Value prev_config = 4;
+
+  // user details
+  UserDetails user_details = 5;
```
aaron-steinfeld, line 60: Now assuming it's the the user making the change this should be better named to indicate this goes with the update timestamp. But it also makes me wonder - should we be storing the creator too to keep these symmetrical?

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -152,12 +158,17 @@ message UpsertAllConfigsRequest {
 
 message UpsertAllConfigsResponse {
   repeated UpsertedConfig upserted_configs = 1;
+  UserDetails user_details = 2;
 
-  message UpsertedConfig {
+  message  UpsertedConfig {
```
aaron-steinfeld, line 163: Stray space

aaron-steinfeld (changes requested):

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
aaron-steinfeld: It needs to work the same as tenant ID. The caller doesn't read the context and write it into the request, the server reads the context and persists it. This is the request object.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
Vaibhav090420: created and updated timestamp is already present in config document

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
Vaibhav090420: Got it, using requestContext to get email id

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
aaron-steinfeld: Right. I'm trying to get at the fact that the pattern is the same here. If we're updating this "user" on each write, it represents the user who last updated it only. If we're only persisting it on create but not update, it represents the user who created it only. 

If we only need one or the other that's fine, but since they're two different concepts (Creator and last updater) we should use the specific terminology vs a more ambiguous just "user"

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -270,8 +271,21 @@ private String getTenantId() {
                 ::asRuntimeException);
   }
 
-  // TODO : get the userId from the context
+  private String getUserEmail() {
+    return RequestContext.CURRENT
```
aaron-steinfeld, line 275: 1. should this be an exception or should we fall back? If we have no email available, then does it make sense to say "config service is not supported"? I don't believe so - it feels like a nice to have to me.
2. If we do throw, make sure to use contextual status exception builder. Also this isn't an argument, it's more of a failed assertion about expected system state. I'd call it INTERNAL.

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
```diff
@@ -77,7 +81,8 @@ public ConfigDocument(
       @JsonProperty(TENANT_ID_FIELD_NAME) String tenantId,
       @JsonProperty(CONTEXT_FIELD_NAME) String context,
       @JsonProperty(VERSION_FIELD_NAME) long configVersion,
-      @JsonProperty(USER_ID_FIELD_NAME) String userId,
+      @JsonProperty(LAST_UPDATED_USER_ID_FIELD_NAME) String lastUpdatedUserId,
```
aaron-steinfeld, line 84: so I certainly prefer this name, but given this is an existing field it makes me a little uncomfortable. Two options:

- if there's no data in there and you test this against some existing entries that it gracefully handles defaulting it/ignoring the old empty string, we can go with this. I _think_ it should but wouldn't want to merge without actually trying. And for that matter, if that causes an issue the new one might anyway.
- otherwise, we can rename the code but leave the actual string value that represents the json field name unchanged.

aaron-steinfeld: Looks like there's both a jackson based and proto based serialization config in here. The proto one should be fine. The jackson one is set to ignore unknown on deserialize so should be fine for the rename
```
new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
Only question is how does it default the new ones? May have to set a default annotation.

aaron-steinfeld: Also, adding a compatibility test would be good. This is unit testable, doesn't need to be manual.

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
```diff
@@ -26,11 +28,14 @@
  */
 @lombok.Value
 @Slf4j
+@Builder
```
aaron-steinfeld, line 31: still needed with the latest change?

aaron-steinfeld (dismissed):
one nit, not blocking

config-service-impl/src/main/java/org/hypertrace/config/service/store/ConfigDocument.java
Vaibhav090420: Not required will remove it

aaron-steinfeld (approved):

PR #214, opened 2024-06-06
chore: add regex capture to label application rule

saxenakshitiz:
Defined regex capture to perform operation on attribute value like substring etc.

### DIFF
--- a/label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
+++ b/label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
@@ -8,6 +8,7 @@ import static org.hypertrace.config.validation.GrpcValidatorUtils.validateReques
 import com.google.protobuf.Message;
 import com.google.re2j.Matcher;
 import com.google.re2j.Pattern;
+import com.google.re2j.PatternSyntaxException;
 import io.grpc.Status;
 import java.util.ArrayList;
 import java.util.List;
@@ -243,9 +244,10 @@ public class LabelApplicationRuleValidatorImpl implements LabelApplicationRuleVa
         .getTokenExtractionRulesList()
         .forEach(
             tokenExtractionRule -> {
-              validKeys.add(tokenExtractionRule.getKey());
               if (tokenExtractionRule.hasAlias()) {
                 validKeys.add(tokenExtractionRule.getAlias());
+              } else {
+                validKeys.add(tokenExtractionRule.getKey());
               }
             });
     Pattern pattern = Pattern.compile("\\{(\\\\}|[^}])*}");
@@ -264,6 +266,15 @@ public class LabelApplicationRuleValidatorImpl implements LabelApplicationRuleVa
   private void validateTokenExtractionRule(
       Action.DynamicLabel.TokenExtractionRule tokenExtractionRule) {
     validateNonDefaultPresenceOrThrow(tokenExtractionRule, tokenExtractionRule.KEY_FIELD_NUMBER);
+    if (tokenExtractionRule.hasRegexCapture()) {
+      try {
+        Pattern.compile(tokenExtractionRule.getRegexCapture());
+      } catch (PatternSyntaxException ex) {
+        throwInvalidArgumentException(
+            String.format(
+                "Invalid regex capture expression: %s", tokenExtractionRule.getRegexCapture()));
+      }
+    }
   }
 
   private void throwInvalidArgumentException(String description) {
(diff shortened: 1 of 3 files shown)


label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -243,9 +244,10 @@ public void validateLabelExpression(Action.DynamicLabel dynamicLabel) {
         .getTokenExtractionRulesList()
         .forEach(
             tokenExtractionRule -> {
-              validKeys.add(tokenExtractionRule.getKey());
               if (tokenExtractionRule.hasAlias()) {
                 validKeys.add(tokenExtractionRule.getAlias());
+              } else {
```
sanket-mundra, line 249: Unintentional change?

saxenakshitiz: Nope. If alias is provided then alias should be used not key.

sanket-mundra: Okay. So, till now we weren't using alias at all?

sanket-mundra (approved):

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
saxenakshitiz: Earlier both alias and key were being used. I will change to use alias if alias is provided.
Please note this expression is not exposed in UI so far, so no one should be using it.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -264,6 +266,15 @@ public void validateLabelExpression(Action.DynamicLabel dynamicLabel) {
   private void validateTokenExtractionRule(
       Action.DynamicLabel.TokenExtractionRule tokenExtractionRule) {
     validateNonDefaultPresenceOrThrow(tokenExtractionRule, tokenExtractionRule.KEY_FIELD_NUMBER);
+    if (tokenExtractionRule.hasRegexCapture()) {
+      try {
+        Pattern.compile(tokenExtractionRule.getRegexCapture());
```
aaron-steinfeld, line 271: I think we have a validator somewhere that validates a number of capture groups - and we should be using re2

aaron-steinfeld: Oh we are :) I forget they use identical names to make it drop in. Anyway, `RegexValidator.validateCaptureGroupCount` is what I was referring to

saxenakshitiz: Done

aaron-steinfeld (approved):

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleValidatorImpl.java
```diff
@@ -264,6 +265,9 @@ public void validateLabelExpression(Action.DynamicLabel dynamicLabel) {
   private void validateTokenExtractionRule(
       Action.DynamicLabel.TokenExtractionRule tokenExtractionRule) {
     validateNonDefaultPresenceOrThrow(tokenExtractionRule, tokenExtractionRule.KEY_FIELD_NUMBER);
+    if (tokenExtractionRule.hasRegexCapture()) {
+      RegexValidator.validateCaptureGroupCount(tokenExtractionRule.getRegexCapture(), 1);
```
sanket-mundra, line 269: Why are we restricting only to one capture group?

saxenakshitiz: Since we need a single value, we can use value only from one group

aaron-steinfeld: You'd create create two rules with one capture group each.

sanket-mundra: But that would lead to two different labels, right? If we want both of of them in a single label?

aaron-steinfeld: No - a dynamic label defines tokens and an expression that indicates how to combine them - this PR is adding a new way to extract a token:
```proto
    message DynamicLabel {
      string label_expression = 1; // expression to combine tokens to generate label, example "${deployment}_${version}"
      repeated TokenExtractionRule token_extraction_rules = 2;
      message TokenExtractionRule {
        string key = 1;
        optional string json_path = 2;
        optional string alias = 3;
        optional string regex_capture = 4;
      }
    }
```

sanket-mundra: Oh, okay

PR #217, opened 2024-06-07
refactor: add support for client timeouts

aaron-steinfeld:
## Description
This change adds support for client timeouts on all config store abstractions, using a default of 5s if unset.


object-store/src/main/java/org/hypertrace/config/objectstore/ClientConfig.java
```diff
@@ -0,0 +1,11 @@
+package org.hypertrace.config.objectstore;
+
+import java.time.Duration;
+import lombok.Value;
+
+@Value
+public class ClientConfig {
+  Duration timeout;
+
+  public static ClientConfig DEFAULT = new ClientConfig(Duration.ofSeconds(5));
```
yogesh-gogia1, line 10: Default looks good, 
max for this as observed in past is <5s

yogesh-gogia1: nit: we can add a TODO here to read this value POJO from config

yogesh-gogia1 (approved):

object-store/src/main/java/org/hypertrace/config/objectstore/ClientConfig.java
aaron-steinfeld: Because this is used by so many different services, I don't love providing it from config - since that means the config itself is needed to build the abstract store object which means every service now needs to get access to it and pass it in. The override case certainly could be from config, but let me think through how/if we could accomplish this.

tim-mwangi (dismissed):

object-store/src/main/java/org/hypertrace/config/objectstore/ClientConfig.java
aaron-steinfeld: Added a todo, and moved the default up to 10s to make sure we're safe (since the intention here is to be as compatible with existing behavior as possible we don't even want to approach the 5s limit)

tim-mwangi (approved):

PR #216, opened 2024-06-07
chore | Proto changes for adding evaluation targets for each exclude span rule

AnandShivansh:
## Description
Please include a summary of the change, motivation and context.

### Testing
Please describe the tests that you ran to verify your changes. Please summarize what did you test and what needs to be tested e.g. deployed and tested helm chart locally. 

### Checklist:
- [ ] My changes generate no new warnings
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] Any dependent changes have been merged and published in downstream modules

### Documentation
Make sure that you have documented corresponding changes in this repository or [hypertrace docs repo](https://github.com/hypertrace/hypertrace-docs-website) if required.

PR #219, opened 2024-06-07
Revert "chore | Proto changes for adding evaluation targets for each exclude span rule"

AnandShivansh:
Reverts hypertrace/config-service#216


aaron-steinfeld (approved):

aaron-steinfeld:
overriding proto revert for unreleased change

PR #226, opened 2024-08-14
ENG-48445: Added generic filter and condition upsert condition (for optimistic updates)

harshit-kumar-v2:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
+++ b/config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
@@ -38,6 +38,9 @@ message UpsertConfigRequest {
   // optional - only required if config applies to a specific context.
   // If empty, specified config is associated with a default context.
   string context = 4;
+
+  //optional - only required when we want to have optimistic locking for update
+  Filter upsertCondition = 5;
 }
 
 message UpsertConfigResponse {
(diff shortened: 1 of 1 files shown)


config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -161,3 +164,39 @@ message UpsertAllConfigsResponse {
...
+
+enum RelationalOperator {
+  RELATIONAL_OPERATOR_UNSPECIFIED = 0;
+  RELATIONAL_OPERATOR_EQ = 1;
+  RELATIONAL_OPERATOR_NEQ = 2;
+  RELATIONAL_OPERATOR_IN = 3;
+  RELATIONAL_OPERATOR_NOT_IN = 4;
+  RELATIONAL_OPERATOR_GT = 5;
+  RELATIONAL_OPERATOR_LT = 6;
+  RELATIONAL_OPERATOR_GE = 7;
```
aaron-steinfeld, line 194: nit - GTE is a bit more readable, assuming this is greater than or equal to. Not trying to save chars here.

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -161,3 +164,39 @@ message UpsertAllConfigsResponse {
...
+
+message Filter {
+  oneof type {
+    RelationalFilter relational_filter = 1;
+    LogicalFilter logical_filter = 2;
+  }
+}
+
+message RelationalFilter {
+  string lhs = 1;
```
aaron-steinfeld, line 176: in general, spell out words. the lhs here is the config_json_path right?
(and operator, value for op, rhs).

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -38,6 +38,9 @@ message UpsertConfigRequest {
   // optional - only required if config applies to a specific context.
   // If empty, specified config is associated with a default context.
   string context = 4;
+
+  //optional - only required when we want to have optimistic locking for update
```
aaron-steinfeld, line 42: Can you describe how this behaves if no config exists (the create side of upsert)? Does it always pass? always fail? And the optimistic locking is more of a use case description, i'd drop it.

```suggestion
  // an optional condition that must pass in order for an update operation to succeed. If the target record does not exist...
```

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
```diff
@@ -161,3 +164,39 @@ message UpsertAllConfigsResponse {
...
+  RELATIONAL_OPERATOR_IN = 3;
+  RELATIONAL_OPERATOR_NOT_IN = 4;
+  RELATIONAL_OPERATOR_GT = 5;
+  RELATIONAL_OPERATOR_LT = 6;
+  RELATIONAL_OPERATOR_GE = 7;
+  RELATIONAL_OPERATOR_LE = 8;
+}
+
+enum LogicalOperator {
+  COMPOSITE_OPERATOR_UNSPECIFIED = 0;
```
aaron-steinfeld, line 199: match enum name

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
harshit-kumar-v2: i am think to have this conditional update only when the Filter field is present (which uses doc store update method)

other case like create and update with no filter can work as it is (current uses doc store upsert method which is deprecated i can use createOrReplace method instead)

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
harshit-kumar-v2: done

aaron-steinfeld (approved):
lgtm

config-service-api/src/main/proto/org/hypertrace/config/service/v1/config_service.proto
aaron-steinfeld: makes sense. Just need to reject the create + filter case with validation

PR #230, opened 2024-08-23
ENG-48723: Filter builder fix and object store changes

harshit-kumar-v2:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/config-service-impl/src/main/java/org/hypertrace/config/service/store/FilterBuilder.java
+++ b/config-service-impl/src/main/java/org/hypertrace/config/service/store/FilterBuilder.java
@@ -2,19 +2,22 @@ package org.hypertrace.config.service.store;
 
 import static org.hypertrace.config.service.store.ConfigDocument.CONFIG_FIELD_NAME;
 
-import com.google.protobuf.ListValue;
-import com.google.protobuf.Struct;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.protobuf.Value;
 import io.grpc.Status;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import lombok.SneakyThrows;
+import org.hypertrace.config.proto.converter.ConfigProtoConverter;
 import org.hypertrace.config.service.v1.LogicalFilter;
 import org.hypertrace.config.service.v1.RelationalFilter;
 import org.hypertrace.core.documentstore.Filter;
 
 class FilterBuilder {
+  private final ObjectMapper objectMapper;
+
+  FilterBuilder() {
+    this.objectMapper = new ObjectMapper();
+  }
 
   public Filter buildDocStoreFilter(org.hypertrace.config.service.v1.Filter filter) {
     switch (filter.getTypeCase()) {
@@ -101,38 +104,9 @@ class FilterBuilder {
     return String.format("%s.%s", CONFIG_FIELD_NAME, configJsonPath);
   }
 
-  public Object convertValueToObject(Value value) {
-    switch (value.getKindCase()) {
-      case NULL_VALUE:
-        return null;
-      case BOOL_VALUE:
-        return value.getBoolValue();
-      case NUMBER_VALUE:
-        return value.getNumberValue();
-      case STRING_VALUE:
-        return value.getStringValue();
-      case STRUCT_VALUE:
-        return convertStructToMap(value.getStructValue());
-      case LIST_VALUE:
-        return convertListValueToList(value.getListValue());
-      default:
-        throw new IllegalArgumentException("Unknown value type: " + value.getKindCase());
-    }
-  }
-
-  private Map<String, Object> convertStructToMap(Struct struct) {
-    Map<String, Object> map = new HashMap<>();
-    for (Map.Entry<String, Value> entry : struct.getFieldsMap().entrySet()) {
-      map.put(entry.getKey(), convertValueToObject(entry.getValue()));
-    }
-    return map;
-  }
-
-  private List<Object> convertListValueToList(ListValue listValue) {
-    List<Object> list = new ArrayList<>();
-    for (Value value : listValue.getValuesList()) {
-      list.add(convertValueToObject(value));
-    }
-    return list;
+  @SneakyThrows
+  private Object convertValueToObject(Value value) {
+    return objectMapper.readValue(
+        ConfigProtoConverter.convertToJsonString(value), new TypeReference<>() {});
   }
 }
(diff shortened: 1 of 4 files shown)


object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
```diff
@@ -153,24 +154,39 @@ public Optional<T> getData(RequestContext context, String id) {
...
+        context,
+        UpsertConfigRequest.newBuilder()
+            .setResourceName(this.resourceName)
+            .setResourceNamespace(this.resourceNamespace)
+            .setContext(this.getContextFromData(data))
+            .setConfig(this.buildValueFromData(data))
+            .build());
+  }
+
+  public ContextualConfigObject<T> upsertObject(RequestContext context, T data, Filter filter) {
```
harshit-kumar-v2, line 178: I thought on this again.

We don't need new abstraction we can just overload the method in base store reasons:
1. The Filter is a generic class can be used directly without a generic type.
2. It was difficult to convert different type of request (like update, or get ) under a single abstract method which builds filter
3. we can build the filter in Impl class and pass the filter.
4. We can have overload on other get methods and this seems simpler for impl class also

harshit-kumar-v2: update dashboard will look like
```
@Override
  public void updateDashboard(
      UpdateDashboardRequest request, StreamObserver<UpdateDashboardResponse> responseObserver) {
    try {
      this.validator.validate(requestContext, request);
      Dashboard existingDashboard =
          this.dashboardStore
              .getData(requestContext, request.getId())
              .orElseThrow(Status.NOT_FOUND::asRuntimeException);

      Dashboard.Builder dashboardBuilder =
          existingDashboard.toBuilder()
              .setName(request.getName())
              .setJson(request.getJson())
              .setLastUpdatedTimestamp(timestampConverter.convert(clock.instant()));

      if (request.hasDescription()) {
        dashboardBuilder.setDescription(request.getDescription());
      }

      Filter upsertCondition =
          Filter.newBuilder()
              .setRelationalFilter(
                  RelationalFilter.newBuilder()
                      .setConfigJsonPath(
                          Dashboard.getDescriptor()
                              .findFieldByNumber(Dashboard.LAST_UPDATED_TIMESTAMP_FIELD_NUMBER)
                              .getJsonName())
                      .setOperator(RelationalOperator.RELATIONAL_OPERATOR_EQ)
                      .setValue(Values.of(Timestamps.toString(request.getPreviousUpdateTimestamp()))))
              .build();

      responseObserver.onNext(
          UpdateDashboardResponse.newBuilder()
              .setDashboard(
                  this.dashboardStore
                      .upsertObject(requestContext, dashboardBuilder.build(), upsertCondition)
                      .getData())
              .build());
    } catch (Exception exception) {
      Exception decoratedException = decorateException(requestContext, exception);
      log.warn(
          "Failed to update dashboard for request: {} with context: {}",
          request,
          requestContext,
          decoratedException);
      responseObserver.onError(exception);
    }
  }
```

config-service-impl/src/main/java/org/hypertrace/config/service/store/FilterBuilder.java
```diff
@@ -106,4 +100,39 @@ private Filter evaluateLeafExpression(RelationalFilter relationalFilter) {
   private String buildConfigFieldPath(String configJsonPath) {
     return String.format("%s.%s", CONFIG_FIELD_NAME, configJsonPath);
   }
+
+  public Object convertValueToObject(Value value) {
```
aaron-steinfeld, line 104: I wonder if we need these / already have them?

In the doc write flow, we have the implementer convert between proto and value. The proto serializes to a json string which is basically what we want here. In other words:
- does this belong in `ConfigProtoConverter` where we already have Value->Proto,  Proto->Value, Value->String, String->Value methods?
- If yes, could we just do Value->String and then toss it in object mapper to convert it like this?

config-service-impl/src/main/java/org/hypertrace/config/service/store/FilterBuilder.java
```diff
@@ -54,47 +61,34 @@ private Filter evaluateCompositeExpression(LogicalFilter logicalFilter) {
   }
 
   private Filter evaluateLeafExpression(RelationalFilter relationalFilter) {
+    Object value = this.convertValueToObject(relationalFilter.getValue());
```
aaron-steinfeld, line 64: So are we sure the Value object doesn't work as-is here? What does doc store do with it?

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
harshit-kumar-v2: i get an idea but not exactly.
Let me try and push few commits for reference

harshit-kumar-v2: > with a few caveats
what are the limitation you see? i will try to see if i can handle them

config-service-impl/src/main/java/org/hypertrace/config/service/store/FilterBuilder.java
aaron-steinfeld: Got it, I had assumed they were stringifying it for some reason (and it stringifies to JSON). In that case, approach above makes sense.

object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
aaron-steinfeld: > what are the limitations you see? i will try to see if i can handle them

Just the points from the other comments. Mainly that 
- we should avoid writing custom conversion code - we can definitely figure something out between the existing converter and/or object mapper.
- the point about making the update w/ filter protected. We can let the impl class deal with building it and expose a typed api. We don't need the impl class exposing the generic filter method without the conversion (i.e. callers to the specific store shouldn't be building a filter)

harshit-kumar-v2: Made the above changes and corresponding config service code: https://github.com/Traceableai/config-service/pull/2055.

> I also think it could be worthwhile to provide some filter building utilities for easily going from a first class proto filter object to a generic one, but we can park that and implement it a few times before pulling out commonalities

on this what type of utilities i can add?

aaron-steinfeld: Conversion looks good. Can you fix the second point - make your new method protected so it doesn't expose it to all callers.

> on this what type of utilities i can add?

Let's come back to this once we build a few implementations. But basically what I was thinking is most typed filters are a series of optional (field, primitive) pairs that are used with equality, and then we AND together any provided pairs. I suspect we could automatically build the generic filter with minimal config (e.g. providing a mapping of filter field number to the json path). But anyway as I said, we can revisit in the future only if it seems worthwhile.

harshit-kumar-v2: right, i know what can we add but at this stage not have enough sets to build one.

aaron-steinfeld (approved):

PR #234, opened 2024-08-29
Fix environment and service name match for unary span filter expression

AnandShivansh:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/validation/SpanProcessingConfigRequestValidator.java
+++ b/span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/validation/SpanProcessingConfigRequestValidator.java
@@ -18,6 +18,8 @@ import org.hypertrace.span.processing.config.service.v1.LogicalSpanFilterExpress
 import org.hypertrace.span.processing.config.service.v1.RelationalSpanFilterExpression;
 import org.hypertrace.span.processing.config.service.v1.SpanFilter;
 import org.hypertrace.span.processing.config.service.v1.SpanFilterValue;
+import org.hypertrace.span.processing.config.service.v1.UnarySpanFilterExpression;
+import org.hypertrace.span.processing.config.service.v1.UnarySpanFilterExpression.LeftOperandCase;
 import org.hypertrace.span.processing.config.service.v1.UpdateExcludeSpanRule;
 import org.hypertrace.span.processing.config.service.v1.UpdateExcludeSpanRuleRequest;
 
@@ -70,6 +72,9 @@ public class SpanProcessingConfigRequestValidator {
       case RELATIONAL_SPAN_FILTER:
         validateRelationalSpanFilter(filter);
         break;
+      case UNARY_SPAN_FILTER:
+        validateUnarySpanFilter(filter);
+        break;
       default:
         throw Status.INVALID_ARGUMENT
             .withDescription("Unexpected filter case: " + printMessage(filter))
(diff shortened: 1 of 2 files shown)


skjindal93 (approved):

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/validation/SpanProcessingConfigRequestValidator.java
```diff
@@ -96,6 +101,19 @@ private void validateRelationalSpanFilter(SpanFilter filter) {
     }
   }
 
+  private void validateUnarySpanFilter(SpanFilter filter) {
+    validateNonDefaultPresenceOrThrow(
+        filter.getUnarySpanFilter(), UnarySpanFilterExpression.OPERATOR_FIELD_NUMBER);
+
+    final LeftOperandCase leftOperandCase = filter.getUnarySpanFilter().getLeftOperandCase();
```
aaron-steinfeld, line 108: I think I'm missing something. You're saying field must be set on 106, then immediately saying if field is set it's also an error?

span-processing-utils/src/main/java/org/hypertrace/config/span/processing/utils/SpanFilterMatcher.java
```diff
@@ -38,18 +45,24 @@ public boolean matchesEnvironment(SpanFilter spanFilter, Optional<String> enviro
...
+    }
     if (spanFilter.hasRelationalSpanFilter()) {
       return matchesServiceName(spanFilter.getRelationalSpanFilter(), serviceName);
     } else {
       if (spanFilter
           .getLogicalSpanFilter()
           .getOperator()
           .equals(LogicalOperator.LOGICAL_OPERATOR_AND)) {
+        if (hasNoRelationalFilters(spanFilter.getLogicalSpanFilter())) {
+          return true;
```
aaron-steinfeld, line 59: This doesn't necessarily hold. It seems like you're making assumptions about structure here.  e.g. `AND(AND(serviceName = x, otherCondition))` would stop searching as soon as it got inside the first AND.

span-processing-utils/src/main/java/org/hypertrace/config/span/processing/utils/SpanFilterMatcher.java
```diff
@@ -59,6 +72,12 @@ public boolean matchesServiceName(SpanFilter spanFilter, String serviceName) {
     }
   }
 
+  private boolean hasNoRelationalFilters(LogicalSpanFilterExpression logicalSpanFilterExpression) {
+    return logicalSpanFilterExpression.getOperandsCount() == 0
```
aaron-steinfeld, line 76: This should recurse. Also the operand count is redundant, since if it was 0 none match would return true.

AnandShivansh: Will fix in the later PR. This is known problem. Environment and servicename is not checked at more than 1 nested levels

span-processing-config-service-impl/src/main/java/org/hypertrace/span/processing/config/service/validation/SpanProcessingConfigRequestValidator.java
aaron-steinfeld: Sorry my mistake here - I saw the word FIELD and thought it was checking for the field.. field. But it's checking for the operator field.

Given this though, recommend removing this from the proto rather than relying on validation since it was just added and not yet in use.

PR #235, opened 2024-08-30
chore: add support for system label application rules

saxenakshitiz:


label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -27,29 +37,42 @@
...
       Channel configChannel, Config config, ConfigChangeEventGenerator configChangeEventGenerator) {
-    int maxDynamicRules = DEFAULT_MAX_DYNAMIC_LABEL_APPLICATION_RULES_PER_TENANT;
-    if (config.hasPath(LABEL_APPLICATION_RULE_CONFIG_SERVICE_CONFIG)) {
-      Config labelApplicationRuleConfig =
-          config.getConfig(LABEL_APPLICATION_RULE_CONFIG_SERVICE_CONFIG);
-      if (labelApplicationRuleConfig.hasPath(MAX_DYNAMIC_LABEL_APPLICATION_RULES_PER_TENANT)) {
-        maxDynamicRules =
-            labelApplicationRuleConfig.getInt(MAX_DYNAMIC_LABEL_APPLICATION_RULES_PER_TENANT);
-      }
+    Config labelApplicationRuleConfig =
```
aaron-steinfeld, line 55: Can we please move all this new code out of here. We don't want to overload business logic into the grpc impls, they're just routers. Please create a new config class that handles loading these values from config.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -27,29 +37,42 @@
...
+        config.hasPath(LABEL_APPLICATION_RULE_CONFIG_SERVICE_CONFIG)
```
aaron-steinfeld, line 56: hasPath checks aren't necessary - just put an empty array as a default value in application.conf (which is less code and makes the config easier to understand anyway). Same goes for the other configs, although they're pre-existing.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -140,6 +173,12 @@ public void deleteLabelApplicationRule(
     try {
       RequestContext requestContext = RequestContext.CURRENT.get();
       this.requestValidator.validateOrThrow(requestContext, request);
+      String labelApplicationRuleId = request.getId();
+      if (systemLabelApplicationRulesMap.containsKey(labelApplicationRuleId)) {
+        // Deleting a system label application rule is not allowed
```
aaron-steinfeld, line 178: We generally allow users to delete default config. Why is this different?

saxenakshitiz: Can be done. However deleting a default rule will require a new config store to persist it. Will add it if need arises.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
saxenakshitiz: Done

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
saxenakshitiz: will do it in separate PR

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -27,29 +30,15 @@
...
-    if (config.hasPath(LABEL_APPLICATION_RULE_CONFIG_SERVICE_CONFIG)) {
-      Config labelApplicationRuleConfig =
-          config.getConfig(LABEL_APPLICATION_RULE_CONFIG_SERVICE_CONFIG);
-      if (labelApplicationRuleConfig.hasPath(MAX_DYNAMIC_LABEL_APPLICATION_RULES_PER_TENANT)) {
-        maxDynamicRules =
-            labelApplicationRuleConfig.getInt(MAX_DYNAMIC_LABEL_APPLICATION_RULES_PER_TENANT);
-      }
-    }
-    this.maxDynamicLabelApplicationRulesAllowed = maxDynamicRules;
+    this.labelApplicationRuleConfig = new LabelApplicationRuleConfig(config);
```
aaron-steinfeld, line 39: Can you make the typed config object a constructor arg. If using that from the factory is a challenge, at least overload it so we can easily test this class with a mock config.

label-application-rule-config-service-impl/src/test/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImplTest.java
```diff
@@ -44,20 +44,29 @@
...
     mockGenericConfigService =
         new MockGenericConfigService().mockUpsert().mockGet().mockGetAll().mockDelete();
     Channel channel = mockGenericConfigService.channel();
     ConfigChangeEventGenerator configChangeEventGenerator = mock(ConfigChangeEventGenerator.class);
-    Config config =
-        ConfigFactory.parseMap(
-            Map.of(
-                LABEL_APPLICATION_RULE_CONFIG_SERVICE_CONFIG,
-                Map.of(MAX_DYNAMIC_LABEL_APPLICATION_RULES_PER_TENANT, 2)));
+    String configStr =
```
aaron-steinfeld, line 59: Likewise here - This doesn't belong in the impl test, we can mock instead and then also add a config class test that passes json and checks the config parse.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
saxenakshitiz: Done

label-application-rule-config-service-impl/src/test/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImplTest.java
saxenakshitiz: Done

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
saxenakshitiz: Keeping it as is. May require changes across multiple application.conf

label-application-rule-config-service-impl/src/test/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImplTest.java
```diff
@@ -44,20 +43,26 @@
 import org.junit.jupiter.api.Test;
 
 public class LabelApplicationRuleConfigServiceImplTest {
+  private static final String SYSTEM_LABEL_APPLICATION_RULE_STR =
```
aaron-steinfeld, line 46: This test is mocking the config, so mock it (i.e. return a prebuilt rule) rather than do the real parse logic and put it behind a mock. Not only is that not the responsibility of this class/test, but if a change were added to the rule structure this test would miss it silently.

saxenakshitiz: Done. Hopefully last review comment to address ;-)

aaron-steinfeld (approved):

PR #236, opened 2024-09-02
chore: add support to delete system label application rule

saxenakshitiz:


aaron-steinfeld (approved):
Breaking change expected due to new proto module

aaron-steinfeld (approved):

PR #249, opened 2024-10-25
Revert "revert guice7 to guice6"

surajpuvvada:
Reverts hypertrace/config-service#248


aaron-steinfeld (changes requested):
The original revert was correct. We do not want guice 7 here.

PR #255, opened 2024-11-20
Use new document store search api

bhubam:
## Description

### Testing

### Checklist:

### Documentation


config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -202,12 +209,11 @@ public Map<ConfigResourceContext, ContextSpecificConfig> getContextConfigs(
   @Override
   public List<ContextSpecificConfig> getAllConfigs(ConfigResource configResource)
       throws IOException {
-    Query query = new Query();
-    query.setFilter(this.getConfigResourceFilter(configResource));
-    query.addOrderBy(new OrderBy(VERSION_FIELD_NAME, false));
+    org.hypertrace.core.documentstore.query.Query query = buildQuery(configResource);
```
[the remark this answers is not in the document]
bhubam: Old api was getting some other places also, moved to new one everywhere

suresh-prakash (approved):

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -94,7 +102,7 @@ public UpsertedConfig writeConfig(
             .asRuntimeException();
       }
     } else {
-      collection.upsert(latestDocKey, latestConfigDocument);
+      collection.createOrReplace(latestDocKey, latestConfigDocument);
```
harshit-kumar-v2, line 105: @suresh-prakash 
This is the change i had made few months back and then reverted cause of issue in downstream doc store api .
It was fixed in https://github.com/hypertrace/document-store/pull/207

but recently we revert this too (cause of issue in Iam service we saw recently) https://github.com/hypertrace/document-store/pull/213

we would want to be sure on this

suresh-prakash: @harshit-kumar-v2 There is no issue with the `createOrReplace` API itself. The problem is only when we have a value that starts with a dollar. That case is not handled and we haven't faced a real use case for it. So, this change is safe.

avinashkolluru (approved):

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
aaron-steinfeld: It's not safe because the values are provided by external users - same thing we discussed last time this went in and was reverted. So we need to revert this again until we can prioritize fixing doc store @suresh-prakash.

aaron-steinfeld: Ah didn't mean to pile on, just came across the already-existing comments on the revert PR! Any further discussion can be had there.

PR #260, opened 2024-12-06
chore: update buf

aaron-steinfeld:


.github/workflows/pr-build.yml
```diff
@@ -47,15 +47,7 @@ jobs:
...
-
-      - name: Lint protos
-        uses: bufbuild/buf-lint-action@v1
-
-      - name: Check for breaking changes
-        uses: bufbuild/buf-breaking-action@v1
-        with:
-          against: '.git#branch=origin/main'
\ No newline at end of file
+          github_token: ${{ github.token }}
```
tim-mwangi, line 53: If you want you can use `setup_only: true`

tim-mwangi (approved):

.github/workflows/pr-build.yml
aaron-steinfeld: This automatically runs all the checks though - why use it as setup only unless we needed something real custom?

tim-mwangi: Yeah. I figured :) I did not need the rest since I was calling buf build directly in make for TPA

PR #267, opened 2025-01-08
Fix rule name duplication check in update

Deepanshu0703:
## Description

### Testing

### Checklist:

### Documentation

### DIFF
--- a/notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleConfigServiceRequestValidator.java
+++ b/notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleConfigServiceRequestValidator.java
@@ -34,7 +34,9 @@ public class NotificationRuleConfigServiceRequestValidator {
     validateRequestContextOrThrow(requestContext);
     validateNonDefaultPresenceOrThrow(request, UpdateNotificationRuleRequest.ID_FIELD_NUMBER);
     validateNonDuplicateNotificationRuleOrThrow(
-        request.getNotificationRuleMutableData().getRuleName(), existingNotificationRules);
+        request.getId(),
+        request.getNotificationRuleMutableData().getRuleName(),
+        existingNotificationRules);
     validateNotificationRuleMutableData(request.getNotificationRuleMutableData());
   }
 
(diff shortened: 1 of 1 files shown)


notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleConfigServiceRequestValidator.java
```diff
@@ -52,6 +54,21 @@ private void validateNonDuplicateNotificationRuleOrThrow(
     }
   }
 
+  private void validateNonDuplicateNotificationRuleOrThrow(
+      String id, String ruleName, List<NotificationRule> existingNotificationRules) {
```
aaron-steinfeld, line 58: Prefer to share the impl. For example mark ID nullable and change the method above to delegate to this one.

Deepanshu0703: Resolved

notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleConfigServiceRequestValidator.java
```diff
@@ -41,23 +43,9 @@ public void validateUpdateNotificationRuleRequest(
...
-            .asRuntimeException();
-      }
-    }
-  }
-
-  private void validateNonDuplicateNotificationRuleOrThrow(
-      String id, String ruleName, List<NotificationRule> existingNotificationRules) {
-    for (NotificationRule existingNotificationRule : existingNotificationRules) {
-      if (!existingNotificationRule.getId().equals(id)
+      if ((Objects.isNull(id) || !existingNotificationRule.getId().equals(id))
```
aaron-steinfeld, line 48: You could put an explicit null check here if you want, but it's not really needed since no existing notification would have a null ID by definition and comparison an existing ID to null is valid. Basically, it doesn't change the logic and makes it a little more confusing to read.

aaron-steinfeld (dismissed):

aaron-steinfeld (approved):

PR #270, opened 2025-01-15
Allow user to update system labels

Vaibhav090420:
## Description

### Testing

### Checklist:

### Documentation


labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/LabelsConfigServiceImpl.java
```diff
@@ -230,19 +195,21 @@ public void updateLabel(
...
       }
       Label oldLabel =
           labelStore
               .getData(requestContext, request.getId())
-              .orElseThrow(Status.NOT_FOUND::asRuntimeException);
+              .orElseGet(
+                  () ->
+                      Optional.ofNullable(
+                              systemLabelInfoUtils
+                                  .getSystemLabelsIdLabelMap(getExistingLabels(requestContext))
```
aman-bansal, line 210: i am confused here? why do you need to find system labels ids from existing labels? We dont change system config ids when we store them in DB

aman-bansal: It should be as simple as

```
Label label =
          labelStore
              .getData(requestContext, labelId)
              .orElseGet(
                  () ->
                      Optional.ofNullable(
                              systemLabelInfoUtils
                                  .getDefaultSystemLabels()
                                  .filter(label -> label.getId().equals(labelId))
                                  .findFirst()
                          .orElseThrow(Status.NOT_FOUND::asRuntimeException));
```

labels-config-service-impl/src/main/java/org/hypertrace/label/config/service/SystemLabelInfoUtils.java
```diff
@@ -0,0 +1,69 @@
...
+  private static final String LABELS_CONFIG_SERVICE_CONFIG = "labels.config.service";
+  private static final String SYSTEM_LABELS = "system.labels";
+
+  private final Config config;
+
+  public SystemLabelInfoUtils(Config config) {
+    this.config = config;
+  }
+
+  public List<Label> getSystemLabels(List<Label> existingLabels) {
```
aman-bansal, line 27: i dont think you would ever need to evaluate system labels from existing labels. same comment for all methods

Vaibhav090420: In case of system label name updation, we would require it to evaluate from the existing labels in db. If we do not want the name updation on system label then, it is not required

labels-config-service-impl/src/test/java/org/hypertrace/label/config/service/LabelsConfigServiceImplTest.java
```diff
@@ -333,15 +334,9 @@ void test_system_updateLabel() {
...
+              .setData(LabelData.newBuilder().setKey(systemLabel.getData().getKey()).build())
               .build();
-      Throwable exception =
-          assertThrows(
-              StatusRuntimeException.class,
-              () -> {
-                labelConfigStub.updateLabel(updateLabelRequest);
-              });
-      assertEquals(Status.INVALID_ARGUMENT, Status.fromThrowable(exception));
+      assertDoesNotThrow(() -> labelConfigStub.updateLabel(updateLabelRequest));
```
aman-bansal, line 339: this test case is wrong. we should add necessary validations

aman-bansal (approved):

skjindal93 (approved):

PR #272, opened 2025-01-20
ENG-55741: filtering support on the get all notification rule api

patelraj0602:


notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleFilteredStore.java
```diff
@@ -52,4 +54,16 @@ protected Value buildValueFromData(NotificationRule object) {
...
     return object.getId();
   }
+
+  @Override
+  protected Optional<NotificationRule> filterConfigData(
+      NotificationRule data, NotificationRuleFilter filter) {
+    return Optional.of(data)
+        .filter(
+            notificationRule ->
+                filter
```
anujgoyal1, line 64: Check a condition for hasFilter.
We don't want to apply any filter if filter is not set or list is empty.

anujgoyal1 (approved):

kotharironak (approved):

PR #277, opened 2025-02-19
ENG-57596 Add get label application rule api

Vaibhav090420:
## Description

### Testing

### Checklist:

### Documentation


label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -87,25 +89,32 @@ public void getLabelApplicationRules(
...
+  }
+
+  @Override
+  public void getLabelApplicationRule(
+      GetLabelApplicationRuleRequest request,
+      StreamObserver<GetLabelApplicationRuleResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      LabelApplicationRule labelApplicationRule =
+          getAllLabelApplicationRules(requestContext).stream()
```
saxenakshitiz, line 111: We don't need to fetch all label application rules in this case. Use getData(...) method in store  impl.

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto
```diff
@@ -13,6 +13,9 @@ service LabelApplicationRuleConfigService {
   // get all label application rules
   rpc GetLabelApplicationRules(GetLabelApplicationRulesRequest) returns (GetLabelApplicationRulesResponse) {}
 
+  // get label application rules
+  rpc GetLabelApplicationRule(GetLabelApplicationRuleRequest) returns (GetLabelApplicationRuleResponse) {}
```
aaron-steinfeld, line 17: A request/response existed but no RPC? Prefer using the existing RPC and just add filter support rather than a second RPC. It's going to be easier anyway, since otherwise you need transformers for both RPCs that confusingly will reference each other.

Vaibhav090420: Why not to have a separate api, do not think it will harm as we are sending all the labels in the getAllApi. I want to use it in GQL for transformer to fetch the labelApplicationRule for update and delete transformer. And if we add a filter in GetAll should it be a single id or repeated. One thing the getAllResponse return a list of labelApplicationRule, given if we give list then we have to main the order of id to rule.

Vaibhav090420: The Transformer for hypertrace are in a different module. But dao module for labels in a different module (There also we have to introduce an method by fetching all the labels  and iterate over them and dependencies on labels modules, adding api seems more efficient)

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -113,6 +115,25 @@ public void getLabelApplicationRules(
...
 
+  @Override
+  public void getLabelApplicationRule(
+      GetLabelApplicationRuleRequest request,
+      StreamObserver<GetLabelApplicationRuleResponse> responseObserver) {
+    try {
+      RequestContext requestContext = RequestContext.CURRENT.get();
+      LabelApplicationRule rule =
+          this.labelApplicationRuleStore
+              .getData(requestContext, request.getId())
```
aaron-steinfeld, line 126: This would also need to respect the deleted system IDs (another benefit from sharing the existing RPC)

label-application-rule-config-service-api/src/main/proto/org/hypertrace/label/application/rule/config/service/v1/label_application_rule_config_service.proto
Vaibhav090420: will go with adding repeated string filter in the GetAllRequest

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -102,10 +103,17 @@ public void getLabelApplicationRules(
               .filter(rule -> !labelApplicationRuleIds.contains(rule.getId()))
               .filter(rule -> !deletedSystemLabelApplicationRuleIds.contains(rule.getId()))
               .collect(Collectors.toUnmodifiableList());
+      List<LabelApplicationRule> allLabelApplicationRules =
+          Stream.concat(
+                  labelApplicationRules.stream(), filteredSystemLabelApplicationRules.stream())
+              .filter(
```
aaron-steinfeld, line 109: Please make `LabelApplicationRuleStore` a `IdentifiedObjectStoreWithFilter` and apply the filter there rather than in the impl method. We try to keep the impls logic-free (this one obviously falls a bit short of that goal, but we don't need to make it worse)

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -102,10 +103,17 @@ public void getLabelApplicationRules(
...
+                  rule ->
+                      !request.hasFilter()
+                          || request.getFilter().getIdsList().contains(rule.getId()))
```
aaron-steinfeld, line 112: prefer to just check if the id list is empty rather than if it has a filter. That way, if a second filter condition is added this logic still holds.

label-application-rule-config-service-impl/src/main/java/org/hypertrace/label/application/rule/config/service/LabelApplicationRuleConfigServiceImpl.java
```diff
@@ -102,10 +103,17 @@ public void getLabelApplicationRules(
               .filter(rule -> !labelApplicationRuleIds.contains(rule.getId()))
               .filter(rule -> !deletedSystemLabelApplicationRuleIds.contains(rule.getId()))
               .collect(Collectors.toUnmodifiableList());
+      List<LabelApplicationRule> allLabelApplicationRules =
```
aaron-steinfeld, line 106: naming: this isn't all rules, just the matching ones.

aaron-steinfeld (approved):

skjindal93 (approved):

PR #281, opened 2025-03-11
support to enable/disable a notif rule

sidharth-jain23:
- added support to enable/disable a notification rule by adding `disabled` field in `NotificationRuleMutableData` message
- added support to optionally exclude disabled rules in bulk fetch request by adding `exclude_disabled` field in `NotificationRuleFilter`
- went with `exclude_disabled` instead of `include_disabled` since the default behaviour required would be to fetch all rules. UI should continue to fetch all rules irrespective of whether disabled or not
- added relevant tests


0001-ENG-58592-support-to-enable-disable-a-notif-rule.patch
```diff
@@ -0,0 +1,186 @@
+From 86dda8a30f2aa3e612d66e4417bd949a62746a87 Mon Sep 17 00:00:00 2001
```
saxenakshitiz, line 1: remove this file

notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleFilteredStore.java
```diff
@@ -67,6 +67,10 @@ protected Optional<NotificationRule> filterConfigData(
                         .contains(
                             notificationRule
                                 .getNotificationRuleMutableData()
-                                .getEventConditionType()));
+                                .getEventConditionType()))
+        .filter(
```
saxenakshitiz, line 71: this filter can be merged with above filter itself?

saxenakshitiz (dismissed):

notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule_config_service.proto
```diff
@@ -37,6 +37,7 @@ message GetAllNotificationRulesRequest {
 
 message NotificationRuleFilter {
   repeated string event_condition_type = 1;
+  bool exclude_disabled = 2;
```
aaron-steinfeld, line 40: the double negative here is confusing. Make it an optional bool for enabled. Unset is all, enabled = true or enabled = false - it also just supports more use cases.

notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleFilteredStore.java
```diff
@@ -61,12 +61,14 @@ protected Optional<NotificationRule> filterConfigData(
...
             notificationRule ->
-                filter.getEventConditionTypeList().isEmpty()
-                    || filter
-                        .getEventConditionTypeList()
-                        .contains(
-                            notificationRule
-                                .getNotificationRuleMutableData()
-                                .getEventConditionType()));
+                (!notificationRule.getNotificationRuleMutableData().getDisabled()
+                        || !filter.getExcludeDisabled())
```
aaron-steinfeld, line 65: the nesting here gets confusing. Suggest breaking out each predicate like:
``` 
satisfiesEnabledPredicate(notificationRule, filter) && 
  satisfiesConditionTypePredicate(notificationRule, filter)
```

notification-rule-config-service-api/src/main/proto/org/hypertrace/notification/config/service/v1/notification_rule_config_service.proto
sidharth-jain23: Looks like for protobuf 3, in java for primitive fields you can't directly check if the field was set because they always have a default value.

aaron-steinfeld: That's the default behavior for a primitive field like `bool` - but that's where the modifier `optional` comes in. https://protobuf.dev/programming-guides/field_presence/#presence-in-proto3-apis

notification-rule-config-service-impl/src/main/java/org/hypertrace/notification/config/service/NotificationRuleFilteredStore.java
```diff
@@ -61,14 +61,21 @@ protected Optional<NotificationRule> filterConfigData(
...
-                (!notificationRule.getNotificationRuleMutableData().getDisabled()
-                        || !filter.getExcludeDisabled())
-                    && (filter.getEventConditionTypeList().isEmpty()
-                        || filter
-                            .getEventConditionTypeList()
-                            .contains(
-                                notificationRule
-                                    .getNotificationRuleMutableData()
-                                    .getEventConditionType())));
+                satisfiesEnabledPredicate(notificationRule, filter)
```
aaron-steinfeld, line 64: actually now that I see it written out, if you want you could even write it as 
```java
 return Optional.of(data)
        .filter(rule -> satisfiesEnabledPredicate(rule, filter))
        .filter(rule -> satisfiesConditionTypePredicate(rule, filter)));
```

Not strictly required, It's all about readability 😉

aaron-steinfeld (approved):

PR #287, opened 2025-04-15
Updated object store with Filtering and Pagination

pavan-traceable:
## Description
Updaed IdentifiedObjectStore With Filtering and Pagination.

### DIFF
--- a/config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
+++ b/config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
@@ -8,12 +8,17 @@ import javax.annotation.Nullable;
 import org.hypertrace.config.service.change.event.api.ConfigChangeEventGenerator;
 import org.hypertrace.config.service.v1.ConfigServiceGrpc;
 import org.hypertrace.config.service.v1.Filter;
+import org.hypertrace.config.service.v1.GetAllConfigsRequest;
 import org.hypertrace.config.service.v1.Pagination;
 import org.hypertrace.config.service.v1.SortBy;
 import org.hypertrace.core.grpcutils.context.RequestContext;
 
 public abstract class IdentifiedFilterPushedDownObjectStore<T, F, S>
     extends IdentifiedObjectStore<T> {
+
+  private final ConfigServiceGrpc.ConfigServiceBlockingStub configServiceBlockingStub;
+  private final String resourceNamespace;
+  private final String resourceName;
 
   protected IdentifiedFilterPushedDownObjectStore(
       ConfigServiceGrpc.ConfigServiceBlockingStub configServiceBlockingStub,
@@ -27,6 +32,9 @@ public abstract class IdentifiedFilterPushedDownObjectStore<T, F, S>
         resourceName,
         configChangeEventGenerator,
         clientConfig);
+    this.configServiceBlockingStub = configServiceBlockingStub;
+    this.resourceNamespace = resourceNamespace;
+    this.resourceName = resourceName;
   }
 
   public List<ContextualConfigObject<T>> getMatchingObjects(
@@ -61,6 +69,40 @@ public abstract class IdentifiedFilterPushedDownObjectStore<T, F, S>
 
   public Optional<T> getMatchingData(RequestContext context, F filterInput, List<S> sortInput) {
     return getMatchingObject(context, filterInput, sortInput).map(ConfigObject::getData);
+  }
+
+  List<ContextualConfigObject<T>> getMatchingObjects(
+      RequestContext context, Filter filter, List<SortBy> sortByList, Pagination pagination) {
+    return context
+        .call(
+            () ->
+                this.configServiceBlockingStub
+                    .withDeadline(getDeadline())
+                    .getAllConfigs(buildGetAllConfigsRequest(filter, sortByList, pagination)))
+        .getContextSpecificConfigsList()
+        .stream()
+        .map(
+            contextSpecificConfig ->
+                ContextualConfigObjectImpl.tryBuild(
+                    contextSpecificConfig, this::buildDataFromValue))
+        .flatMap(Optional::stream)
+        .collect(
+            Collectors.collectingAndThen(
+                Collectors.toUnmodifiableList(), this::orderFetchedObjects));
+  }
+
+  private GetAllConfigsRequest buildGetAllConfigsRequest(
+      Filter filter, List<SortBy> sortByList, Pagination pagination) {
+    GetAllConfigsRequest.Builder getAllConfigsRequest =
+        GetAllConfigsRequest.newBuilder()
+            .setResourceName(this.resourceName)
+            .setResourceNamespace(this.resourceNamespace)
+            .setFilter(filter)
+            .addAllSortBy(sortByList);
+    if (pagination != null) {
+      getAllConfigsRequest.setPagination(pagination);
+    }
+    return getAllConfigsRequest.build();
   }
 
   protected abstract SortBy buildSort(S sortInput);
--- a/config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
+++ b/config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
@@ -38,9 +38,7 @@ public abstract class IdentifiedFilterPushedDownObjectStore<T, F, S>
 
   public List<ContextualConfigObject<T>> getMatchingObjects(
       RequestContext context, F filterInput, List<S> sortInput) {
-    Filter filter = buildFilter(filterInput);
-    List<SortBy> sortByList = sortInput.stream().map(this::buildSort).collect(Collectors.toList());
-    return getMatchingObjects(context, filter, sortByList, null);
+    return getMatchingObjects(context, filterInput, sortInput, null);
   }
 
   public List<T> getMatchingData(
--- a/config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
+++ b/config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
@@ -7,12 +7,17 @@ import java.util.stream.Collectors;
 import org.hypertrace.config.service.change.event.api.ConfigChangeEventGenerator;
 import org.hypertrace.config.service.v1.ConfigServiceGrpc;
 import org.hypertrace.config.service.v1.Filter;
+import org.hypertrace.config.service.v1.GetAllConfigsRequest;
 import org.hypertrace.config.service.v1.Pagination;
 import org.hypertrace.config.service.v1.SortBy;
 import org.hypertrace.core.grpcutils.context.RequestContext;
 
 public abstract class IdentifiedFilterPushedDownObjectStore<T, F, S>
     extends IdentifiedObjectStore<T> {
+
+  private final ConfigServiceGrpc.ConfigServiceBlockingStub configServiceBlockingStub;
+  private final String resourceName;
+  private final String resourceNamespace;
 
   protected IdentifiedFilterPushedDownObjectStore(
       ConfigServiceGrpc.ConfigServiceBlockingStub configServiceBlockingStub,
@@ -26,13 +31,23 @@ public abstract class IdentifiedFilterPushedDownObjectStore<T, F, S>
         resourceName,
         configChangeEventGenerator,
         clientConfig);
+    this.configServiceBlockingStub = configServiceBlockingStub;
+    this.resourceName = resourceName;
+    this.resourceNamespace = resourceNamespace;
   }
 
   public List<ContextualConfigObject<T>> getMatchingObjects(
       RequestContext context, F filterInput, List<S> sortInput, Pagination pagination) {
     Filter filter = buildFilter(filterInput);
     List<SortBy> sortByList = sortInput.stream().map(this::buildSort).collect(Collectors.toList());
-    return getMatchingObjectsWithFilter(context, filter, pagination, sortByList);
+    return getMatchingObjectsWithFilter(context, filter, sortByList, pagination);
+  }
+
+  public List<ContextualConfigObject<T>> getMatchingObjects(
+      RequestContext context, F filterInput, List<S> sortInput) {
+    Filter filter = buildFilter(filterInput);
+    List<SortBy> sortByList = sortInput.stream().map(this::buildSort).collect(Collectors.toList());
+    return getMatchingObjectsWithFilter(context, filter, sortByList, null);
   }
 
   public List<T> getMatchingData(
@@ -44,8 +59,7 @@ public abstract class IdentifiedFilterPushedDownObjectStore<T, F, S>
 
   public Optional<ContextualConfigObject<T>> getMatchingObject(
       RequestContext context, F filterInput, List<S> sortInput) {
-    List<ContextualConfigObject<T>> results =
-        getMatchingObjects(context, filterInput, sortInput, Pagination.getDefaultInstance());
+    List<ContextualConfigObject<T>> results = getMatchingObjects(context, filterInput, sortInput);
     if (results.size() > 1) {
       throw Status.FAILED_PRECONDITION
           .withDescription("Multiple objects found when expecting at most one")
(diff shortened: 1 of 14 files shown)


config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -0,0 +1,42 @@
...
+      RequestContext context, F filterInput, Pagination pagination) {
+    Filter filter = buildFilter(filterInput);
+    List<SortBy> sortByList = buildSortByList(filterInput);
+    return getAllObjectsWithFilter(context, filter, pagination, sortByList);
+  }
+
+  protected abstract Filter buildFilter(F filterInput);
+
+  /** Constructs a list of SortBy from client filter or returns an empty list. */
+  protected List<SortBy> buildSortByList(F filterInput) {
```
aaron-steinfeld, line 39: - Sorts should be passed into the getAllObjects call. I would probably abstract it and let the callers specify, similar to filter. So a new type param `S` and accept a `List<S>` with a method like `protected abstract SortBy buildSort(S sortInput);`
- We'd also want to add equivalent implementations for all the other methods in `IdentifiedObjectStoreWithFilter` there's a plural and singular version of both get object and get data.
- Since it's a new class we can also fix the naming, it's not "getAll" since we're applying a filter, it should be "getMatching"

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
```diff
@@ -114,6 +116,33 @@ public List<ContextualConfigObject<T>> getAllObjects(RequestContext context) {
                 Collectors.toUnmodifiableList(), this::orderFetchedObjects));
   }
 
+  public List<ContextualConfigObject<T>> getAllObjectsWithFilter(
```
aaron-steinfeld, line 119: This is specific to the push down impl and should be a private method there. We don't want to expose the internals to the consumers of the object store

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: Fixed all of them, please check now.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
pavan-traceable: Updated to package private

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -26,17 +27,43 @@ protected IdentifiedFilterPushedDownObjectStore(
...
+    return filterConfigData(configObject.getData(), filter)
+        .map(filteredData -> updateConfigData(configObject, filteredData));
+  }
+
+  private ContextualConfigObject<T> updateConfigData(
+      ContextualConfigObject<T> configObject, T updatedData) {
+    return ((ContextualConfigObjectImpl) configObject).toBuilder().data(updatedData).build();
+  }
+
+  protected abstract Optional<T> filterConfigData(T data, F filter);
```
aaron-steinfeld, line 64: We don't need this one or L53, since we've replaced it with the pushed down filter

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -26,17 +27,43 @@ protected IdentifiedFilterPushedDownObjectStore(
...
+  public List<T> getMatchingConfigData(
+      RequestContext context, F filterInput, List<S> sortInput, Pagination pagination) {
+    return getMatchingObjects(context, filterInput, sortInput, pagination).stream()
+        .map(ConfigObject::getData)
+        .collect(Collectors.toUnmodifiableList());
+  }
+
+  public Optional<ContextualConfigObject<T>> getObject(
+      RequestContext context, String id, F filter) {
+    return getObject(context, id).flatMap(configObject -> filterObject(configObject, filter));
```
aaron-steinfeld, line 46: Hmm so I missed here that we have two parts of a filter - the id and the actual filter object. We should be able to turn the ID into a `Filter` and then logically and them together, that way we can avoid requiring a separate in memory implementation, removing L53-64 (and more importantly, removing one abstract method to simplify implementers).

If we can't internal convert ID into a filter (because it probably depends on the specifics of T, now that I think about it) then a couple other options would be to either 1. add an abstract method to build an ID filter or 2. just remove ID from this signature and leave it to the implementer to add support for an ID filter as part of `F` if they have this use case. In essence, this just becomes a singular get that should return an empty optional if no results, an optional if one result, and throw if more than one result.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -26,17 +27,43 @@ protected IdentifiedFilterPushedDownObjectStore(
...
+  private ContextualConfigObject<T> updateConfigData(
+      ContextualConfigObject<T> configObject, T updatedData) {
+    return ((ContextualConfigObjectImpl) configObject).toBuilder().data(updatedData).build();
+  }
+
+  protected abstract Optional<T> filterConfigData(T data, F filter);
+
+  protected abstract Filter buildFilter(F filterInput);
+
+  protected abstract List<SortBy> buildSort(List<S> sortInput);
```
aaron-steinfeld, line 68: nit - this implementation can deal with the list part of the translation, we can simplify.

```suggestion
  protected abstract SortBy buildSort(S sortInput);
```

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
aaron-steinfeld: But why put it in the parent class to begin with? Is some other code calling it? If not, we should make it fully private it and move it into the filter impl object store.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: Fixed it

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: Ok

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: I hope i have addressed all these concerns, please check.

pavan-traceable: One doubt, what if id is not in config object in MongoDB. How does filter work since we always config as prefix

pavan-traceable: One more issue is there is already filter in client which might not have id, in that case how to handle it ? I am working on apispecs to use this model. please check api_spec.proto which has ApiSpecFilter

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
```diff
@@ -114,6 +116,33 @@ public List<ContextualConfigObject<T>> getAllObjects(RequestContext context) {
                 Collectors.toUnmodifiableList(), this::orderFetchedObjects));
   }
 
+  List<ContextualConfigObject<T>> getMatchingObjectsWithFilter(
```
aaron-steinfeld, line 119: Raising a new thread in case the other comment got missed - https://github.com/hypertrace/config-service/pull/287#discussion_r2045120709

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -0,0 +1,64 @@
...
+      RequestContext context, F filterInput, List<S> sortInput, Pagination pagination) {
+    return getMatchingObjects(context, filterInput, sortInput, pagination).stream()
+        .map(ConfigObject::getData)
+        .collect(Collectors.toUnmodifiableList());
+  }
+
+  public Optional<ContextualConfigObject<T>> getMatchingObject(
+      RequestContext context, F filterInput, List<S> sortInput) {
+    List<ContextualConfigObject<T>> results =
+        getMatchingObjects(context, filterInput, sortInput, Pagination.getDefaultInstance());
```
aaron-steinfeld, line 48: as soon as you see code with `getDefaultInstance()` we should be thinking "antipattern!". If we don't want pagination, don't use it. Explicitly assigning a default value to indicate "unset" breaks presence checking and even if it works today that relies on how downstream code is written which makes it really easy to introduce bugs later since you're breaking the protobuf contract.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
aaron-steinfeld: > One doubt, what if id is not in config object in MongoDB. How does filter work since we always config as prefix

ID must be in the config object since that's what the `Identified` indicates in the parent class here. The implementer must provide a definition for `abstract String getContextFromData(T data)`. 

> One more issue is there is already filter in client which might not have id, in that case how to handle it ? I am working on apispecs to use this model. please check api_spec.proto which has ApiSpecFilter

Then they would add ID to the filter if they want it to be pushed down. We typically want these by-id filters anyway, so in this model not only do we get the push-down support but now we're also exposing more functionality in the RPC.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedObjectStore.java
pavan-traceable: Fixed it

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: Updated

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -58,6 +72,40 @@ public Optional<T> getMatchingData(RequestContext context, F filterInput, List<S
...
+                ContextualConfigObjectImpl.tryBuild(
+                    contextSpecificConfig, this::buildDataFromValue))
+        .flatMap(Optional::stream)
+        .collect(
+            Collectors.collectingAndThen(
+                Collectors.toUnmodifiableList(), this::orderFetchedObjects));
+  }
+
+  private GetAllConfigsRequest buildGetAllConfigsRequest(
+      Filter filter, List<SortBy> sortByList, Pagination pagination) {
```
aaron-steinfeld, line 96: nit - annotate nullability (especially where these same values show up on the public part of the API) since we're inconsistent here, pagination can be null, filter + sort by list can't be.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -26,13 +31,23 @@ protected IdentifiedFilterPushedDownObjectStore(
...
     List<SortBy> sortByList = sortInput.stream().map(this::buildSort).collect(Collectors.toList());
-    return getMatchingObjectsWithFilter(context, filter, pagination, sortByList);
+    return getMatchingObjectsWithFilter(context, filter, sortByList, pagination);
+  }
+
+  public List<ContextualConfigObject<T>> getMatchingObjects(
+      RequestContext context, F filterInput, List<S> sortInput) {
+    Filter filter = buildFilter(filterInput);
+    List<SortBy> sortByList = sortInput.stream().map(this::buildSort).collect(Collectors.toList());
+    return getMatchingObjectsWithFilter(context, filter, sortByList, null);
```
aaron-steinfeld, line 50:
```suggestion
    return getMatchingObjects(context, filter, sortByList, null);
```

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -26,13 +31,23 @@ protected IdentifiedFilterPushedDownObjectStore(
         resourceName,
         configChangeEventGenerator,
         clientConfig);
+    this.configServiceBlockingStub = configServiceBlockingStub;
+    this.resourceName = resourceName;
```
aaron-steinfeld, line 35: So this is why you initially had that method it in the `IdentifiedObjectStore` I'm guessing. Rather than shadowing these variables in the subclass, could we just set them to package private in `IdentifiedObjectStore`?

pavan-traceable: yes, may be i should have explained to you then itself.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: done

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: done

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
aaron-steinfeld: Revisit please - the two overloads don't both need to convert filter + sort. Just have the one without sort call the one with sort as shown in the suggestion

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
aaron-steinfeld: I wish we did this in every method, but at a minimum every method in this callstack that has inconsistent treatment for different arguments.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: Updated. I thought of renaming the method in parent class.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: Updated in all the methods where we use pagination.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
aaron-steinfeld: Sorry I should be clearer with my comments - by setting "them" package private I meant the declarations of the variables from `IdentifiedObjectStore` that you were shadowing in this subclass. 

I would still keep the filter impl in this subclass because that's actually the method doing what this class is built/named for -  pushing down the filter.

pavan-traceable: I have kept methods here and added variables also here. Make variables package private in `IdentifiedObjectStore` leading to errors in ContextuallyIdentifiedObjectStore.
`Cannot reference 'IdentifiedObjectStore.configServiceBlockingStub' before superclass constructor is called`

aaron-steinfeld: That's just a name clash since that inner class is referencing unrelated parent class variables
image

aaron-steinfeld (approved):

PR #288, opened 2025-04-17
Add optional total_count for config service api's

pavan-traceable:
## Description
Add total_count attribute in GetAllConfigsAPI


config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
aaron-steinfeld, line 1: make sure to merge main

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -115,17 +115,15 @@ public void getConfig(
...
                   request.getResourceName(), request.getResourceNamespace(), getTenantId()),
               request.getFilter(),
               request.getPagination(),
-              request.getSortByList());
-      responseObserver.onNext(
-          GetAllConfigsResponse.newBuilder()
-              .addAllContextSpecificConfigs(contextSpecificConfigList)
-              .build());
+              request.getSortByList(),
+              request.getIncludeTotal());
```
aaron-steinfeld, line 125: So within the service, I would consider splitting up the total query - it doesn't really share anything with the config part. That way we're just adding a new API to the configStore, not adding branching and booleans that change behavior

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -233,7 +245,20 @@ public List<ContextSpecificConfig> getAllConfigs(
 
     configList.sort(
```
aaron-steinfeld, line 246: this is a bug, right? It'll discard the sort applied in mongo. I thought we discussed removing the in memory sort in the impl PR but I don't think I followed up on it

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: done

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
pavan-traceable: Fixed.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -230,12 +230,26 @@ public List<ContextSpecificConfig> getAllConfigs(
         processDocument(documentIterator.next(), seenContexts, configList);
       }
     }
-
-    configList.sort(
-        Comparator.comparingLong(ContextSpecificConfig::getCreationTimestamp).reversed());
+    // When there is no requested sort by, sort by creation timestamp default.
```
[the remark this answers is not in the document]
pavan-traceable: Yea, confused with multiple sorts here, i think i have fixed it now.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -115,20 +118,49 @@ public void getConfig(
...
+                      request.getSortByList());
+                } catch (IOException e) {
+                  throw new CompletionException(e);
+                }
+              });
+
+      CompletableFuture<Long> totalCountFuture = null;
+      if (request.getIncludeTotal()) {
+        totalCountFuture =
+            CompletableFuture.supplyAsync(
```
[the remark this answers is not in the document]
pavan-traceable: done

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -115,20 +118,49 @@ public void getConfig(
...
+          CompletableFuture.supplyAsync(
+              () -> {
+                try {
+                  return configStore.getAllConfigs(
+                      configResource,
+                      request.getFilter(),
+                      request.getPagination(),
+                      request.getSortByList());
+                } catch (IOException e) {
+                  throw new CompletionException(e);
```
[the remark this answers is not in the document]
pavan-traceable: Please check it is fixed now.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -115,20 +118,49 @@ public void getConfig(
...
+
+      CompletableFuture<Long> totalCountFuture = null;
+      if (request.getIncludeTotal()) {
+        totalCountFuture =
+            CompletableFuture.supplyAsync(
+                () -> configStore.getMatchingConfigsCount(configResource, request.getFilter()));
+      }
+
+      // Wait for both futures to complete
+      List<ContextSpecificConfig> configList = configListFuture.join();
```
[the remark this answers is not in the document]
pavan-traceable: done

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -115,21 +128,55 @@ public void getConfig(
...
       responseObserver.onCompleted();
-    } catch (Exception e) {
-      log.error("Get all configs failed for request:{}", request, e);
+    } catch (StatusRuntimeException e) {
+      log.error("Get all configs failed for request: {}", request, e);
       responseObserver.onError(e);
+    } catch (Exception e) {
+      log.error("Get all configs failed for request: {}", request, e);
+      responseObserver.onError(
+          Status.INTERNAL.withCause(e).withDescription("Unexpected error").asRuntimeException());
```
aaron-steinfeld, line 179: nit - can use Status.fromThrowable which will propagate a status exception or wrap it in unknown if needed. Whatever the mechanism, avoiding multiple catch clause would be best.

```suggestion
    } catch (Exception e) {
      log.error("Get all configs failed for request: {}", request, e);
      responseObserver.onError(Status.fromThrowable(e));
    }
```

pavan-traceable: done

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -42,6 +46,15 @@ public class ConfigServiceGrpcImpl extends ConfigServiceGrpc.ConfigServiceImplBa
   private static String DEFAULT_USER_ID = "Unknown";
   private static String DEFAULT_USER_EMAIL = "Unknown";
   private final ConfigStore configStore;
+  private static final Executor configExecutor =
+      Executors.newFixedThreadPool(
+          4,
+          runnable -> {
```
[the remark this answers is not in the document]
pavan-traceable: Fixed

pavan-traceable:
@aaron-steinfeld  Added api in IdentifiedFilterPushedDownObjectStore to expose total Count as result.

config-object-store/src/main/java/org/hypertrace/config/objectstore/ConfigsResponseImpl.java
```diff
@@ -0,0 +1,24 @@
+package org.hypertrace.config.objectstore;
+
+import java.util.List;
+
+public class ConfigsResponseImpl<T> implements ConfigsResponse<T> {
```
aaron-steinfeld, line 5: nit: when separating interface and impl, make the impl package private
```suggestion
class ConfigsResponseImpl<T> implements ConfigsResponse<T> {
```

config-object-store/src/main/java/org/hypertrace/config/objectstore/ConfigsResponse.java
```diff
@@ -0,0 +1,9 @@
+package org.hypertrace.config.objectstore;
+
+import java.util.List;
+
+public interface ConfigsResponse<T> {
+  List<ContextualConfigObject<T>> getContextualConfigObjects();
```
aaron-steinfeld, line 6: Do we want to use `List<ContextualConfigObject<T>>` or `List<T>` ? Even though the latter is derivable from `ContextualConfigObject`, the vast majority of our use cases are after `T` directly so by returning this we're introducing an extra mapping for every consumer. Suggest return `List<T>` and we can always add the more complete/verbose form if a use case comes along.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -71,14 +72,41 @@ public Optional<T> getMatchingData(RequestContext context, F filterInput, List<S
     return getMatchingObject(context, filterInput, sortInput).map(ConfigObject::getData);
   }
 
+  public ConfigsResponse<T> getMatchingObjectsWithTotalCount(
+      RequestContext context, F filterInput, List<S> sortInput, Pagination pagination) {
+    List<SortBy> sortByList = sortInput.stream().map(this::buildSort).collect(Collectors.toList());
```
aaron-steinfeld, line 77: nit - can we refactor this to share the implementation with `getMatchingObjects`? other than sending the flag/reading the total on response, they're identical.

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
```diff
@@ -170,13 +165,9 @@ public void getAllConfigs(
...
       responseObserver.onNext(responseBuilder.build());
       responseObserver.onCompleted();
-    } catch (StatusRuntimeException e) {
-      log.error("Get all configs failed for request: {}", request, e);
-      responseObserver.onError(e);
     } catch (Exception e) {
       log.error("Get all configs failed for request: {}", request, e);
-      responseObserver.onError(
-          Status.INTERNAL.withCause(e).withDescription("Unexpected error").asRuntimeException());
+      responseObserver.onError(Status.fromThrowable(e).withCause(e).asRuntimeException());
```
aaron-steinfeld, line 170: fromThrowable already sets the cause if necessary

```suggestion
      responseObserver.onError(Status.fromThrowable(e).asRuntimeException());
```

config-object-store/src/main/java/org/hypertrace/config/objectstore/ConfigsResponse.java
pavan-traceable: Updated

config-service-impl/src/main/java/org/hypertrace/config/service/ConfigServiceGrpcImpl.java
pavan-traceable: fixed

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -72,19 +74,29 @@ public Optional<T> getMatchingData(RequestContext context, F filterInput, List<S
     return getMatchingObject(context, filterInput, sortInput).map(ConfigObject::getData);
   }
 
-  public ConfigsResponse<T> getMatchingObjectsWithTotalCount(
-      RequestContext context, F filterInput, List<S> sortInput, Pagination pagination) {
+  public ConfigsResponse<ContextualConfigObject<T>> getMatchingObjectsWithTotalCount(
+      RequestContext context, F filterInput, List<S> sortInput, @Nullable Pagination pagination) {
+    Filter filter = buildFilter(filterInput);
```
aaron-steinfeld, line 79: The filter and sort conversion can move inside the shared method too I believe

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
```diff
@@ -72,19 +74,29 @@ public Optional<T> getMatchingData(RequestContext context, F filterInput, List<S
     return getMatchingObject(context, filterInput, sortInput).map(ConfigObject::getData);
   }
 
-  public ConfigsResponse<T> getMatchingObjectsWithTotalCount(
-      RequestContext context, F filterInput, List<S> sortInput, Pagination pagination) {
+  public ConfigsResponse<ContextualConfigObject<T>> getMatchingObjectsWithTotalCount(
```
aaron-steinfeld, line 77: So in my earlier comment I was suggesting something more like
```suggestion
  public ConfigsResponse<T> getMatchingDataWithTotalCount(
```

Unless we actually need the `ContextualConfigObject` today, but those verbose versions of the APIs aren't really used much. We can always add the `ConfigsResponse<ContextualConfigObject<T>> getMatchingObjectsWithTotalCount` form if needed later to have the symmetry with the other APs.

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: fixed it

config-object-store/src/main/java/org/hypertrace/config/objectstore/IdentifiedFilterPushedDownObjectStore.java
pavan-traceable: It will be 
`(ConfigsResponse<T>) getMatchingObjects(context, filterInput, sortInput, pagination, true);`

aaron-steinfeld: It shouldn't require a cast - `getMatchingObjects` should be returning `ConfigsResponse<T>` instead of `ConfigsResponse<ContextualConfigObject<T>>` as it's currently declared. The impl would need to change.

pavan-traceable: Understood now, added both objects and data api and updated accordingly

aaron-steinfeld (approved):

PR #289, opened 2025-04-22
Bugfix user filters 

pavan-traceable:
## Description
Fixed the bug around user filters in document config store


config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
```diff
@@ -245,7 +245,7 @@ public long getMatchingConfigsCount(
     return collection.count(query);
   }
 
-  private Query buildQuery(
+  Query buildQuery(
```
aaron-steinfeld, line 248: nit - don't expose a method just to test it. Test it through the public interface.

pavan-traceable: It's bit involved to write cases with public interface, can take it up as next PR.

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
```diff
@@ -358,6 +365,107 @@ void writeAllConfigs() throws IOException {
...
+    org.hypertrace.config.service.v1.SortBy sortBy =
+        org.hypertrace.config.service.v1.SortBy.newBuilder()
+            .setSelection(selection)
+            .setSortOrder(SortOrder.SORT_ORDER_DESC)
+            .build();
+    List<org.hypertrace.config.service.v1.SortBy> sortByList = Collections.singletonList(sortBy);
+
+    Query query = configStore.buildQuery(configResource, filter, pagination, sortByList);
+
+    assertEquals(2, query.getSorts().size());
```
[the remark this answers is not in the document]
aaron-steinfeld: this also removes the need to compare the size and applies to each test. The main benefit is that failures show up as a list diff (e.g. "expected [x, y], got []") rather than a more opaque failure like "expected 2, got 0"

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
aaron-steinfeld: Next PR is fine.

As far as what to change, one option would be to extract a query builder class, in which case testing that makes sense in a unit test (while tests on the store would start mocking it). But an even easier option... this test could just call `getAllConfigs` instead - it passes the same 4 arguments through to the query builder directly. Then just verify the call to `collection.query`.

aaron-steinfeld (approved):
Disabled auto merge (so you can review the comments before deciding whether to merge - please do not use automerge until code has been approved), but if you want to take the comments in a follow up that's fine.

config-service-impl/src/main/java/org/hypertrace/config/service/store/DocumentConfigStore.java
pavan-traceable: I know this, since i need to wait for one more day to get it merged, i thought of doing in next PR. Anyway i have addressed all the comments. We should also take decision to remove sorting on ConfigVersion, i don't its needed any more.

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
```diff
@@ -368,41 +369,75 @@ void writeAllConfigs() throws IOException {
...
+    // Mock the collection.query to return the mock iterator
+    when(collection.query(any(Query.class), any(QueryOptions.class))).thenReturn(mockIterator);
 
-    assertEquals(10, query.getPagination().get().getOffset());
-    assertEquals(20, query.getPagination().get().getLimit());
+    // Act
+    configStore.getAllConfigs(configResource, filter, pagination, sortByList);
+
+    // Capture the Query object passed to collection.query
+    ArgumentCaptor<Query> queryCaptor = ArgumentCaptor.forClass(Query.class);
```
aaron-steinfeld, line 427: arg captors aren't needed here (they're almost always overkill). Let me post a branch to illustrate what I was suggesting

aaron-steinfeld: https://github.com/hypertrace/config-service/pull/290

If that looks good to you, feel free to merge it into your branch (or just take bits and pieces)

pavan-traceable: I have merged your PR. Thanks for that.

aaron-steinfeld (approved):

PR #290, opened 2025-04-23
test: simplify tests

aaron-steinfeld:


config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
```diff
@@ -72,17 +79,19 @@ class DocumentConfigStoreTest {
...
   private static Value config2 = getConfig2();
   private static ConfigResourceContext configResourceContext = getConfigResourceContext();
-  private static Collection collection;
-  private DocumentConfigStore configStore;
-  private Clock mockClock;
-  private FilterBuilder filterBuilder;
 
-  @BeforeEach()
+  private Collection collection;
+  @Mock private Clock mockClock;
```
aaron-steinfeld, line 84: So I added the mockito extension just as a best practice, it's not directly related to the change. But it does lend readability as we need less setup and it will fail if we start mocking things that the test path doesn't touch (you can see some unused mocks I removed in `WriteConfigForUpdateWithUpsertCondition`)

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
```diff
@@ -402,47 +383,39 @@ void buildQuery_withDefaultPagination() throws Exception {
...
                 org.hypertrace.core.documentstore.expression.operators.SortOrder.DESC));
-    assertEquals(expectedSorts, capturedQuery.getSorts());
+
+    configStore.getAllConfigs(
+        new ConfigResource(RESOURCE_NAMESPACE, RESOURCE_NAME, "tenant1"),
+        Filter.getDefaultInstance(),
+        Pagination.getDefaultInstance(),
+        emptyList());
+
+    verify(collection)
```
aaron-steinfeld, line 393: This is what I was suggesting in the other PR - arg captors aren't needed, just call the public API and verify that the query looks like you want it to.

config-service-impl/src/test/java/org/hypertrace/config/service/store/DocumentConfigStoreTest.java
```diff
@@ -577,41 +540,4 @@ public Document next() {
       return documentIterator.next();
     }
   }
-
-  public static class MockDatastore implements Datastore {
```
aaron-steinfeld, line 581: This was overkill, a regular mockito mock can do this more simply.

PR #296, opened 2025-05-13
Proto changes for notification rule delete/update

Ohiduz:


suresh-prakash (dismissed):

skjindal93 (approved):

aaron-steinfeld (commented):
Why are we introducing soft delete for this one config and no others? Can you elaborate more on why this change is needed and what makes notification rules special? @suresh-prakash

